PR #104, opened 2024-10-01
feat: added adjusted and retrofitted gas bills

### LINKED ISSUE #26: feat: ensure there are two gas bills (one for retrofitted or newbuild house, one for non)
- `gasBillYearly` also hasn't been forecasted by CPI yet

zz-hh-aa:
# What does this PR do?
- Adds `gasBillAdjustedYearly` and `gasBillRetrofitYearly` to the `Household` class 
    - They are now included in the `Household` constructor, and also have their own calculation functions `calculateGasBillAdjustedYearly` and `calculateGasBillRetrofitYearly`
- Updates the `Lifetime` class to include `gasBillAdjustedYearly` and `gasBillRetrofitYearly`, inflating their values year on year
    - `interface LifetimeParams` is updated, as is `interface LifetimeData` and the `calculateLifetime` function itself 
- Passes the gas bills to the graphs
    - `gasBillAdjustedYearly` to `LifetimeMarketPurchaseWrapper` and `LifetimeMarketRentWrapper`
    - `gasBillRetrofitYearly` to `LifetimeFairholdLandPurchaseWrapper` and `LifetimeFairholdLandRentWrapper`
  

# Why?
- The previous `gasBillYearly` value we were getting was one static figure for the average house, assuming heating demand of 13,600 kwh. Since the calculator needs to be able to account for different house sizes, types _and_ insulation, one value wasn't sufficient. 
- `gasBillAdjustedYearly` takes the price per kwh and uses the `gasBillAverageYearly` (retried from database), `property.size` and `property.houseType` to estimate a more accurate version of bills for the existing property 
    - `gasBillRetrofitYearly` does the same but with lower kwh demands per m2

Closes #26

### DIFF
--- a/app/components/graphs/LifetimeFairholdLandPurchaseWrapper.tsx
+++ b/app/components/graphs/LifetimeFairholdLandPurchaseWrapper.tsx
@@ -29,7 +29,7 @@ const LifetimeFairholdLandPurchaseWrapper: React.FC<Props> = ({
           incomeThreshold: item.affordabilityThresholdIncome,
           landCost: item.fairholdLandMortgageYearly,
           houseCost: item.depreciatedHouseMortgageYearly,
-          // billsCost: item.gasBillYearly,
+          billsCost: item.gasBillRetrofitYearly,
           maintenanceCost: item.maintenanceCost,
         };
       }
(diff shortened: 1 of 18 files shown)


zz-hh-aa:
Made a mess of this in the rebase / merge conflict fixing; since we have new data soon and needed to replace the math anyway, I'm just going to close this PR.

PR #115, opened 2024-10-11
feat: Create backup script

DafyddLlyr:
## What does this PR do?
 - Creates a script in `package.json` to create a local backup of the Vercel db
 - Runs in docker so it's OS agnostic
 - I've added a copy of the backup to the OSL GDrive as an insurance policy, and I have one locally
 - Please try it out!

## Changes to `.env` files
 - `.env.local` is now used for the local database and Prisma client
 - `.env.production` is used for the prod database (backups)
 - Lets catch up next week to talk this through (with a diagram)


zz-hh-aa:
Thanks Daf--I tried `npm run backup` but encountered this error:
```
> fairhold-dashboard@0.1.0 backup
> ./scripts/backup.sh

'.' is not recognized as an internal or external command,
operable program or batch file.
```

Not quite sure what the issue is--I tried it again, leaving out the `./` in `./scripts...` but got the same thing (`'scripts' is not recognized as an internal or external command, operable program or batch file.`

Any ideas what the issue might be?

zz-hh-aa:
Ah-ha, Windows strikes again... do you think you could update with `cross-env` or `shx` and add to the script eg `"cross-env ./scripts/backup.sh"`? I guess we'll need this for the `test:coverage` script too.

zz-hh-aa:
Alas no @DafyddLlyr 
``` 
Processing fstab with mount -a failed.
Failed to mount C:\, see dmesg for more details.

<3>WSL (9) ERROR: CreateProcessEntryCommon:334: getpwuid(0) failed 2
<3>WSL (9) ERROR: CreateProcessEntryCommon:505: execvpe /bin/bash failed 2
<3>WSL (9) ERROR: CreateProcessEntryCommon:508: Create process not expected to return
```

DafyddLlyr:
Thanks for testing - I'll take a look at using a docker container for this. I think we'll need one for a local dev db anyway. PR should be open again next week!

zz-hh-aa:
This worked for me now, but I did have to change `.env.local` to `.env`, which I guess means we'll also have to update this line in `load-env-and-prisma.js`: `require("dotenv").config({ path: ".env.local" });`

DafyddLlyr:
Closed in favour of https://github.com/theopensystemslab/fairhold-dashboard/pull/141

PR #117, opened 2024-10-15
chore: fix typo in mortgage 'principle'

zz-hh-aa:
Does what it says--fixes instances of mortgage `principal` to `principle`.

### DIFF
--- a/app/models/Mortgage.test.ts
+++ b/app/models/Mortgage.test.ts
@@ -18,7 +18,7 @@ it("correctly calculates the amount of the mortgage ", () => {
     initialDeposit: 0.1,
   });
 
-  expect(mortgage.principal).toBeCloseTo(90);
+  expect(mortgage.principle).toBeCloseTo(90);
 });
 
 it("correctly calculates the amount of monthly payment ", () => {
--- a/app/models/Mortgage.ts
+++ b/app/models/Mortgage.ts
@@ -31,7 +31,7 @@ export class Mortgage {
   /**
    * The principle is the value of the property, minus the deposit
    */
-  principal: number;
+  principle: number;
   monthlyPayment: number;
   totalMortgageCost: number;
   yearlyPaymentBreakdown: MortgageBreakdown;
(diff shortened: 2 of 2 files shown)


DafyddLlyr (commented):
Very good spot thanks - there's a mismatch between a number of functions and the class property.

I'm not sure this is a spelling issue, or UK vs US spellings issue, though - principle and principal are two different words and I think "principal" is right here. 

Principal in a mortgage sense is the amount borrowed, but I think is often confused with the banking term "mortgage in principle" (i.e. in theory you can borrow based on what we know, as in "from first principles").

zz-hh-aa:
Excellent spot! I think you're right and I got wires crossed with mortgage in principle, in any case all consistent to 'principal' now.

app/models/Mortgage.ts
```diff
@@ -53,7 +53,7 @@ export class Mortgage {
     this.yearlyPaymentBreakdown = this.calculateYearlyPaymentBreakdown();
   }
 
-  private calculateMortgagePrinciple() {
+  private calculateMortgageprincipal() {
```
DafyddLlyr, line 56:
```suggestion
  private calculateMortgagePrincipal() {
```

app/models/Mortgage.test.ts
```diff
@@ -18,7 +18,7 @@ it("correctly calculates the amount of the mortgage ", () => {
     initialDeposit: 0.1,
   });
 
-  expect(mortgage.principal).toBeCloseTo(90);
+  expect(mortgage.principle).toBeCloseTo(90);
```
DafyddLlyr, line 21: I think we want this to be `principal` also right?

DafyddLlyr (changes requested):

app/models/Mortgage.test.ts
zz-hh-aa: Yes I think so! 

Thanks for the catches, maybe my find-and-replace-all didn't quite work 🤔

DafyddLlyr (approved):

PR #130, opened 2024-10-24
feat: new maintenance selector

zz-hh-aa:
Adds a radio button selector to the calculator form to select between maintenance spend levels. This PR has been rebased to create the selector with ShadCN.

### DIFF
--- a/app/components/ui/RadioButton.tsx
+++ b/app/components/ui/RadioButton.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { UseFormRegister, Path } from "react-hook-form";
 import { Calculation } from "@/app/schemas/calculationSchema";
 
-interface radioButtonProps {
+interface RadioButtonProps {
   id: string;
   value: string;
   label: string;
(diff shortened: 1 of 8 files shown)


app/components/ui/RadioButton.tsx
```diff
@@ -1,11 +1,12 @@
 import React from "react";
-import { UseFormRegister } from "react-hook-form";
+import { UseFormRegister, Path } from "react-hook-form";
 import { Calculation } from "@/app/schemas/calculationSchema";
 
 interface radioButtonProps {
```
DafyddLlyr, line 5:
```suggestion
interface RadioButtonProps {
```

app/models/constants.ts
```diff
@@ -30,3 +30,5 @@ export const NATIONAL_AVERAGES: nationalAverageType = {
   propertyValue: 49750,
   earningsWeekly: 316.4,
 };
+
+export const MAINTENANCE_LEVELS = ["0.015", "0.02", "0.0375"] as const;
```
DafyddLlyr, line 34: Please see comment on the other PR - I think these should be numbers.

PR #134, opened 2024-10-25
feat: maintenance calculations

### LINKED ISSUE #133: Add input for maintenance

zz-hh-aa:
_This PR uses #130 as base (that handles the UI, this handles the calculations)._

# What does this PR do?
- Updates instances of `maintenanceSpend` to `maintenancePercentage` for clarity
- Adds `maintenancePercentage` to relevant schemas and tests
- Creates a new function `createForecastParameters()` to overwrite `DEFAULT_FORECAST_PARAMETERS` with user input maintenance spend percentage

Closes #133.


app/models/ForecastParameters.ts
```diff
@@ -9,11 +9,27 @@ export interface ForecastParameters {
...
   yearsForecast: 40, // 40 years
   affordabilityThresholdIncomePercentage: 0.35, // percentage of income to afford rent or purchase
 } as const;
+
+/**
+ * Converts maintenance spend string to number and creates forecast parameters
+ * @param maintenancePercentage - Maintenance spend value from form ("0.015" | "0.02" | "0.0375")
+ * @returns ForecastParameters with updated maintenance cost
+ */
+export function createForecastParameters(maintenancePercentage: string): ForecastParameters {
```
DafyddLlyr, line 26: I think there's probably scope to simplify a fair few things in this PR by making sure that `maintenancePercentage` (and `MAINTENANCE_LEVELS`) are numbers at all times. 

Even if it's a string value in the input, we can use Zod to refine it to a number for submission, and then keep it as a number going forward in out codebase.

zz-hh-aa: Ah great catch, I think I've just pushed a fix for this now.

DafyddLlyr (approved):

PR #135, opened 2024-10-25
UI interface using Shadcn components

gabrielegranello:
# What does this PR do?
Instead of using our input components for the UI, it takes advantage of the [Shadcn](https://ui.shadcn.com) library.

# What else has changed ?
Not sure if this was the best approach but, I have introduced a new schema called `formSchema`. This is because typescript was flagging an inconsistency with the `postcode`, and specifically the fact that the `postcode` string was transformed into a `ValidPostcode` via the refine method in `zod` while the input field only accepts `strings`.

Therefore I have introduced a `formSchema`, where zod performs a form validation and still checks the postcode to be valid BUT it does not transform it into a postcode object.

When the data arrives to the API, then the string is transformed into a postcode object by using the actual `CalculationSchema`


zz-hh-aa:
Thanks Gabry, looks great to me! I'll update my maintenance PR (#130) with the ShadCN components too.

zz-hh-aa (approved):

PR #141, opened 2024-11-04
dp/backup restore

DafyddLlyr:
- **feat: Create backup script**
- **fix: Explicity call bash for non-unix environments**
- **feat: Alternative approach using docker**
- **fix: Handle issue with env files**
- **feat: Local dev db, backup and restore scripts, via Docker**
- **docs: Add mermaid diagram to readme**


load-env-and-prisma.js
DafyddLlyr, line 1: I've removed this in favour of calling `dotenv-cli` in the `package.json` scripts.

Prisma docs: https://www.prisma.io/docs/orm/more/development-environment/environment-variables/managing-env-files-and-setting-variables#using-dotenv-cli-via-command-line

zz-hh-aa (approved):
Super exciting, thanks Daf!

Everything works for me locally, though I did run into some errors when running the scripts--I did have to update the end of line sequences from CRLF to LF. Do you know if there's a more OS-agnostic way of handling this?

DafyddLlyr:
Ah good catch @zz-hh-aa  - we can fix this with a `.gitattributes` files. I'll open a PR to handle this 👍

zz-hh-aa:
Worked great--thanks Daf 🙌

PR #145, opened 2024-11-22
feat: update Fairhold formula with new linear one

zz-hh-aa:
# What does this PR do?
- Changes old cosine formula into updated linear one
    - Hardcoded the parameters in the formula as I don't think we will be changing them or need them to be saved as constants externally (which we did before)
- Updates the tests accordingly

# Why?
- For the sake of simplicity, Alastair updated the Fairhold discount formula.

### DIFF
--- a/app/models/Fairhold.ts
+++ b/app/models/Fairhold.ts
@@ -1,50 +1,22 @@
 import * as math from "mathjs";
 
-const DEFAULT_AMPLITUDE = 0.25;
-const DEFAULT_LENGTH = 1;
-const DEFAULT_POSITION = 0.45;
-const DEFAULT_PLATEAU = 0.15;
-const DEFAULT_THRESHOLD = 0.5;
-
-type ConstructorParams = Pick<Fairhold, "affordability" | "landPriceOrRent"> &
-  Partial<
-    Pick<
-      Fairhold,
-      "amplitude" | "length" | "position" | "plateau" | "threshold"
-    >
-  >;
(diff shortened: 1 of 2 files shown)


app/models/Fairhold.ts
```diff
@@ -1,50 +1,22 @@
...
 
   private calculateFairholdDiscount() {
-    if (this.affordability > this.threshold) return this.plateau;
-
-    const discountLand =
-      this.amplitude *
-        math.cos((2 * math.pi * this.affordability) / this.length) +
-      this.position;
-
+    const discountLand = math.max(-1.1 * this.affordability + 0.7, 0.15)
```
DafyddLlyr, line 19: I think it would still be worth making these magic numbers constants, just for documentation and readability.

I agree if they're not variables, we don't need to pass them into the class or have them as properties.

DafyddLlyr (changes requested):

app/models/Fairhold.ts
zz-hh-aa: Makes sense, forgot about that whole readability thing! New commit should take care of this.

app/models/Fairhold.ts
```diff
@@ -16,7 +21,7 @@ export class Fairhold {
   }
 
   private calculateFairholdDiscount() {
-    const discountLand = math.max(-1.1 * this.affordability + 0.7, 0.15)
+    const discountLand = math.max(MULTIPLIER * this.affordability + OFFSET, PLATEAU)
```
DafyddLlyr, line 24: Much nicer!

DafyddLlyr (approved):

PR #146, opened 2024-11-22
feat: Setup Prisma migrations

DafyddLlyr:
## What does this PR do?
 - Bumps Prisma version
 - Sets up migrations, following steps outlined here - https://www.prisma.io/docs/orm/prisma-migrate/getting-started
 - Sets up GH action to automatically run migrations when a new migration is added

Next PR (once this is merged) will test this with a small change 👍


zz-hh-aa (approved):
All makes sense to me! One question I have is if we need `pkey`s to be `id` in each instance (I think `id` is typically the csv index, and perhaps we could have `pkey`s to be actual values from the table? Beyond the scope of this PR but something I was wondering about generally with regards to data shape.

PR #147, opened 2024-11-22
feat: update depreciation 

### LINKED ISSUE #154: Calculate depreciating house value

zz-hh-aa:
# What does this PR do?
`constants.ts`
- Creates new types for calculating depreciation
    - `componentBreakdownType` breaks each component down into the percentage of the house it makes up, how much of its original value it depreciates each year, and what percentage of yearly maintenance spend it accounts for 
    - `houseBreakdownType` is an object with a component breakdown object per-component
- `HOUSE_BREAKDOWN_PERCENTAGES` is exported as a constant for use in `calculateDepreciatedBuildPrice()`

`Property.ts`
- New `calculateComponentValue()` method (separated this out for ease of testing)
- Updates `calculateDepreciatedBuildPrice()`
    - Loops through each object in `HOUSE_BREAKDOWN_PERCENTAGES` and runs `calculateComponentValue()` to calculate depreciated value per-component
    - Sums each value

`Property.test.ts`
- Creates tests for newbuild, depreciated building, component breakdowns

Tests now pass! I ended up opting for a different method where stuff gets calculated again in the test file. Felt a bit more meaningful than just passing the resultant number as the expected value. 

Closes #154

### DIFF
--- a/app/models/Property.test.ts
+++ b/app/models/Property.test.ts
@@ -1,6 +1,5 @@
 import { Property } from "./Property";
-import { HOUSE_BREAKDOWN_PERCENTAGES, MAINTENANCE_LEVELS } from './constants';
-
+import { HOUSE_BREAKDOWN_PERCENTAGES, MAINTENANCE_LEVELS, HouseBreakdown } from './constants';
 describe('Property', () => {
   beforeEach(() => {
     property = new Property({
@@ -26,7 +25,6 @@ describe('Property', () => {
     expect(property.newBuildPrice).toBeCloseTo(186560);
   });
 
-  // 
   it("correctly calculates the bedWeightedAveragePrice", () => {
     expect(property.bedWeightedAveragePrice).toBeCloseTo(219135);
   });
@@ -40,7 +38,8 @@ describe('Property', () => {
   });
   
   it("correctly calculates the values even for number of bedroooms exceeding the max ", () => {
-    property = new Property({
+    // Any bedroom size bigger than 6 beds should be treated as 6 beds, so this compares one instantiated with 20 beds with one instantiated with 6
+    const propertyBig = new Property({
       postcode: "WV8 1HG",
       houseType: "T",
       numberOfBedrooms: 20,
@@ -51,7 +50,20 @@ describe('Property', () => {
       averageMarketPrice: 218091.58,
       itl3: "TLG24",
     });
-    expect(property).toBeInstanceOf(Property);
+
+    const propertySmall = new Property({
+      postcode: "WV8 1HG",
+      houseType: "T",
+      numberOfBedrooms: 6,
+      age: 11,
+      size: 88,
+      maintenancePercentage: 0.02,
+      newBuildPricePerMetre: 2120,
+      averageMarketPrice: 218091.58,
+      itl3: "TLG24",
+    });
+
+    expect(propertyBig.bedWeightedAveragePrice).toEqual(propertySmall.bedWeightedAveragePrice);
   });
 
   it("correctly returns newBuildPrice if newbuild", () => {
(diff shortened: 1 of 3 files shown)


gabrielegranello (approved):
That looks good to me @zz-hh-aa , great work.

app/models/Property.test.ts
```diff
@@ -1,56 +1,183 @@
...
+      expect(result.depreciatedComponentValue).toBeGreaterThanOrEqual(0);
+    });
+
+    it('should calculate correct depreciation for a 10-year-old house', () => {
+      // Calculate the expected value manually
+      const newBuildPrice = property.newBuildPrice;
+      let expectedDepreciatedPrice = 0;
+
+      // Calculate for each component (mimicking calculateComponentValue())
+      for (const [key, value] of Object.entries(HOUSE_BREAKDOWN_PERCENTAGES)) {
```
gabrielegranello, line 164: I just have one question here. Are you basically implementing the same formulas as they are in the class definition correct? Is it still worth comparing with the Python values?

gabrielegranello: Makes sense thanks for explaining.

app/models/Property.test.ts
```diff
@@ -1,56 +1,183 @@
...
+      const result = property.calculateComponentValue(
+        'foundations',
+        property.newBuildPrice,
+        property.age,
+        MAINTENANCE_LEVELS[0]
+      );
+
+      const expectedNewComponentValue = 
+        property.newBuildPrice * 
+        HOUSE_BREAKDOWN_PERCENTAGES.foundations.percentageOfHouse;
```
DafyddLlyr, line 85: I'm personally don't think that this approach of repeating our logic within our tests is the right one. This might be one of the rare places where a magic number is correct.

This approach - 
 - Is more complex to write
 - Is more complex to maintain (any changes need to be updated in many places)
 - Allows us to make a mistake in logic in the tests and class, but having a passing unit test
 
 This sort of sums it up I think - https://aaronsaray.com/2021/do-not-calculate-answers-to-tests/

app/models/Property.test.ts
```diff
@@ -1,56 +1,183 @@
...
+    it("ensures depreciatedComponentValue never goes below 0", () => {
+      const result = property.calculateComponentValue(
+        'ventilationServices',
+        property.newBuildPrice,
+        100, // High age to test possible negative values
+        MAINTENANCE_LEVELS[0]
+      );
+
+      expect(result.depreciatedComponentValue).toBeGreaterThanOrEqual(0);
+    });
```
DafyddLlyr, line 156: In tests like this we can use something like `test.each()` ([docs](https://jestjs.io/docs/api#testeachtablename-fn-timeout))

It might be more meaningful to have the same test, with a number of variable ages, to prove we have a floor at 0.

app/models/Property.test.ts
```diff
@@ -1,56 +1,183 @@
...
+      const expectedMaintenanceAddition = 
+        MAINTENANCE_LEVELS[0] * 
+        property.newBuildPrice * 
+        property.age * 
+        HOUSE_BREAKDOWN_PERCENTAGES.electricalAppliances.percentOfMaintenanceYearly;
+
+      expect(result.maintenanceAddition).toBe(expectedMaintenanceAddition);
+    });
+
+    it("correctly calculates depreciatedComponentValue for ventilation services", () => {
```
DafyddLlyr, line 122: See above comment re:duplicating logic - at the level of complexity it feels very tricky to have duplicated logic here and in the class.

app/models/Property.ts
```diff
@@ -63,7 +69,7 @@ export class Property {
     this.postcode = params.postcode;
     this.houseType = params.houseType;
     this.numberOfBedrooms = params.numberOfBedrooms;
-    this.age = params.age;
+    this.age = params.age - 1; // Subtract 1 because years should be indexed to 0
```
DafyddLlyr, line 72: Helpful comment! This feels a bit tricky though as "age" indicates a non-zero indexed figure.

Should we just return `newBuildPrice` if the age is 0 or 1?

app/models/Property.ts
```diff
@@ -81,18 +87,62 @@ export class Property {
...
+    const maintenanceAddition = 
+      (componentKey === 'foundations' || componentKey === 'structureEnvelope') 
+        ? 0 
+        : maintenanceLevel * newBuildPrice * age * component.percentOfMaintenanceYearly;
+    
+    // Calculate final value
+    let depreciatedComponentValue = 
+      (newComponentValue * depreciationFactor) + maintenanceAddition;
+
+    depreciatedComponentValue < 0 ? depreciatedComponentValue = 0 : depreciatedComponentValue
```
DafyddLlyr, line 138:
```suggestion
    // Do not depreciate below 0
    if (depreciatedComponentValue < 0) depreciatedComponentValue = 0
```

Maybe a bit clearer?

app/models/constants.ts
```diff
@@ -32,3 +32,114 @@ export const NATIONAL_AVERAGES: nationalAverageType = {
...
+export type componentBreakdownType = {
+  /** Component value as percentage of total house value */
+  percentageOfHouse: number, 
+  /** Percentage of the component's total value that is written-down yearly */
+  depreciationPercentageYearly: number,
+  /** Percentage of the yearly maintenance spend allocated to component */
+  percentOfMaintenanceYearly: number
+}
+
+export type houseBreakdownType = {
```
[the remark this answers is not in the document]
DafyddLlyr: It might be simpler to make a union of the keys, and then define this as as record.

```ts
type Component = "foundations" | "structureEnvelope" | ... 
type HouseBreakdown = Record<Component, ComponentBreakdown>
```

app/models/constants.ts
```diff
@@ -32,3 +32,114 @@ export const NATIONAL_AVERAGES: nationalAverageType = {
...
+  waterAndHeatingServices: componentBreakdownType,
+  floorCoverings: componentBreakdownType,
+  landscaping: componentBreakdownType,
+}
+
+/** 
+ * Object (key = component, value = another object with `percentageOfHouse`, `deprecationPercentageYearly` and `percentOfMaintenanceYearly`). Contains a breakdown of a house by component, and the percentage of total value it constitutes:
+ * foundations 21%, structure 25%, cladding 4%, roofing 4%, windows 4%, internal linings 4%, bathroom 4%, fitout 5%,
+ * kitchen units 4%, electrical appliances 4%, electrical services 4%, ventilation services 4%, water and heating services 8%,
+ * floor coverings 2%, landscaping 3%*/
```
DafyddLlyr, line 68: It might be worth encoding these as variables also?

zz-hh-aa: Thanks for the keen eyes!

Do you mean encoding as variables in addition to storing them inside the objects? For instance, instead of:
``` 
export const HOUSE_BREAKDOWN_PERCENTAGES: houseBreakdownType = {
  foundations: {
    percentageOfHouse: .21,
    depreciationPercentageYearly: 0,
    percentOfMaintenanceYearly: 0
  },
  ```
  We might have something like
  ``` 
const foundationsPercent = .21
  export const HOUSE_BREAKDOWN_PERCENTAGES: houseBreakdownType = {
  foundations: {
    percentageOfHouse: foundationsPercent,
...
```
?

app/models/Property.test.ts
```diff
@@ -1,56 +1,183 @@
...
-    maintenancePercentage: 0.02,
-    newBuildPricePerMetre: 2120,
-    averageMarketPrice: 218091.58,
-    itl3: "TLG24",
-  });
-  expect(property).toBeInstanceOf(Property);
-});
+describe('Property', () => {
+  beforeEach(() => {
+    property = new Property({
```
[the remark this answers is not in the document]
zz-hh-aa: I think my thinking was that we would need to re-instantiate after tests changed the parameters of the object, eg in the test lines 41-53:
``` 
   property = new Property({
      postcode: "WV8 1HG",
      houseType: "T",
      numberOfBedrooms: 20,
      age: 11,
      size: 88,
      maintenancePercentage: 0.02,
      newBuildPricePerMetre: 2120,
      averageMarketPrice: 218091.58,
      itl3: "TLG24",
    });
```

However, I think it might be good to rewrite all tests after `oz/test-helpers` (#151) is merged. Is this worth fixing here?

app/models/Property.ts
zz-hh-aa: Yep, that's what's going on
```    
if (this.age === 0) return this.newBuildPrice; // If newbuild, return newBuildPrice and don't depreciate
```

I think it got a bit confusing because we are using 1 as a stand-in for newbuilds in the 'age' form field (I have now made a note to change that), whereas some of the lifetime forecasting math wanted `y = 0` for the first year (eg if appreciating by something like `inflatedPrice = (1 + inflation) ^ y`).

Does that make sense?? Not sure if I've explained myself well here!

app/models/Property.ts
zz-hh-aa: Much clearer 😅 thanks!

app/models/Property.test.ts
zz-hh-aa: Cheers for pointing this out! 
> “I’m using different mechanisms to calculate the response values, so the chance both of those are wrong is very small.” You’re right, except you’re not.

😅

The argument here makes sense to me, I'm just struggling with how to come up with that "known answer" in the first place. If the number is complicated to arrive at, how else will you arrive at it if not through code? I don't actually _know_ what number I expect to come out of the code 🤔 I'll have a go tomorrow at refactoring all of the tests and trying to come up with another way of generating the expected values.

app/models/constants.ts
DafyddLlyr: Yeah this is a tricky one! I think for now the best way to go is just to add `as const` to the variable as well.

app/models/constants.ts
```diff
@@ -25,10 +25,118 @@ export type nationalAverageType = {
...
+    percentageOfHouse: .02, 
+    depreciationPercentageYearly: .039,
+    percentOfMaintenanceYearly: .037
+  },
+  landscaping: {
+    percentageOfHouse: .03, 
+    depreciationPercentageYearly: .0343,
+    percentOfMaintenanceYearly: .056
+  },
+}
```
DafyddLlyr, line 142:
```suggestion
} as const;
```

DafyddLlyr (approved):
@zz-hh-aa Really sorry for the holdup here! All looking good :)

PR #148, opened 2024-11-27
docs: improve tsdoc comments for documentation clarity

zz-hh-aa:
# What's new in this PR
Updated TSDoc across all modules in the `models` directory. To view docs in browser, run `npm run docs` (script is not new). 

# Why
Each time I return to this repo to work on a new feature I find that I've forgotten what a variable name means, why a property is present in a class, or similar--and then I have to re-learn my way around! Hopefully this makes that process easier for all involved. 

Feedback, eg if anything is still unclear, is very welcome.

### DIFF
--- a/app/models/Fairhold.ts
+++ b/app/models/Fairhold.ts
@@ -7,10 +7,25 @@ const PLATEAU = 0.15
 
 type ConstructorParams = Pick<Fairhold, "affordability" | "landPriceOrRent">;
 
+/** The Fairhold class is a generic class 
+ * (meaning it will calculate a discount regardless if it is given a purchase price or monthly rent).
+ * It is instantiated twice,
+ * once each for `FairholdLandPurchase` and `FairholdLandRent`
+ */
 export class Fairhold {
+  /** Affordability is calculated as monthly housing cost / GDHI */
   public affordability: number;
+  /** Depending on whether calculating `FairholdLandPurchase` or `FairholdLandRent`,
+   * pass the relevant figure (open market residual land value or estimated portion of monthly market rent
+   * that goes towards location)
+   */
   public landPriceOrRent: number;
+  /** The Fairhold discount is a multiplier, so Fairhold prices are `discountLand * open market value`
+   */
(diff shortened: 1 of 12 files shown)


app/models/Household.ts
```diff
@@ -9,6 +9,7 @@ import { ForecastParameters } from "./ForecastParameters";
 import { socialRentAdjustmentTypes } from "../data/socialRentAdjustmentsRepo";
 import { Lifetime, LifetimeParams } from "./Lifetime"; 
 
+/** Assumed number of heads per-house */
```
DafyddLlyr, line 12: I had no idea this is what this referred to 💡 

Maybe it's worth changing to a more meaningful name?

app/models/Fairhold.ts
```diff
@@ -7,10 +7,25 @@ const PLATEAU = 0.15
 
 type ConstructorParams = Pick<Fairhold, "affordability" | "landPriceOrRent">;
 
+/** The Fairhold class is a generic class 
+ * (meaning it will calculate a discount regardless if it is given a purchase price or monthly rent).
```
DafyddLlyr, line 11:
```suggestion
/** 
 * The Fairhold class is a generic class 
 * (meaning it will calculate a discount regardless if it is given a purchase price or monthly rent).
```

nit: JSDocs in blocks should have just `/**` as the first line. It's not broken and it does work, but as it's breaking conventions it feels a bit harder to read / unexpected. This applies multiple times throughout the PR below also.

`/** This is the right notation for a single line though 👍  */`

Docs: https://jsdoc.app/about-getting-started

DafyddLlyr (changes requested):
Two small comments - really fab to see more documentation though, super helpful!

app/models/Fairhold.ts
zz-hh-aa: Thanks for letting me know, super useful for future reference! Have now fixed this, I believe :)

DafyddLlyr (approved):

PR #151, opened 2024-11-28
test: test helper functions

zz-hh-aa:
# What's new in this PR
A `testHelpers.ts` file, which contains:
- helper functions that instantiate all relevant classes with simplified values
- TSdoc for each function

# Why
When writing a test for the `Lifetime` class, I realised that it needs all four tenure classes as well as the `Property` class, which contain sub-classes too. Not only was it too messy to instantiate all of that in the `Lifetime.test.ts` file, but I realised that we were creating many instances of these classes and their sub-classes across the test files, and thought it would be straightforward for all testing to have a set of helper functions that instantiate them with default values.

### DIFF
--- a/app/models/Fairhold.ts
+++ b/app/models/Fairhold.ts
@@ -7,10 +7,25 @@ const PLATEAU = 0.15
 
 type ConstructorParams = Pick<Fairhold, "affordability" | "landPriceOrRent">;
 
+/** The Fairhold class is a generic class 
+ * (meaning it will calculate a discount regardless if it is given a purchase price or monthly rent).
+ * It is instantiated twice,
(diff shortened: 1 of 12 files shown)


zz-hh-aa:
Following dev call today, just noting that Claude suggested we go with this route because 
- we are dealing with complex class instances and these functions will instantiate actual instances of our production code
- using functions means we won't accidentally stub behaviour that we want to test
- if class interfaces change, TypeScript will catch errors

zz-hh-aa:
This branches from `oz/resale-values` which branches from `oz/depreciation-update` (where the test issue should be resolved).

DafyddLlyr (approved):

PR #157, opened 2024-12-05
chore: calculate `landToTotalRatio` once and pass it to relevant classes

### LINKED ISSUE #150: Calculate `landToTotalRatio` once
And pass it to all relevant places--currently being calculated multiple times

zz-hh-aa:
# What's new in this PR?
- Removes `averagePrice` and `landPrice` from tenure classes `MarketRent` and `FairholdLandRent` because they were only there to be used to calculate `landToTotalRatio`, which gets calculated in Property` anyway
- Pass `landToTotalRatio` to `MarketRent` and `FairholdLandRent`
- Update tests accordingly (to fit the new params shape)

# Why?
- `landToTotalRatio` was being calculated multiple times and could just be calculated once and passed where needed.

Closes #150

### DIFF
--- a/app/models/Household.ts
+++ b/app/models/Household.ts
@@ -65,11 +65,10 @@ export class Household {
 
     const marketRent = new MarketRent({
       averageRentYearly: averageRentYearly,
-      averagePrice: this.property.averageMarketPrice,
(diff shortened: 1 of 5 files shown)


DafyddLlyr (approved):
Nice!

PR #160, opened 2024-12-06
test(ci): Test Prisma migrations

DafyddLlyr:
## What does this PR do?
- Changes the name of a column to test migrations on CI
- This needed to be done using the `--create-only` flag - without this Prisma would drop and recreate the column. I needed to manually write the correct SQL in the migration file.
- Adds basic doc to explain process

image


package.json
```diff
@@ -13,10 +13,11 @@
     "test": "jest",
     "test:watch": "jest --watch",
     "test:coverage": "jest; open ./coverage/lcov-report/index.html",
-    "prisma:pull": "dotenv -e .env.local npx prisma db pull",
+    "prisma:pull": "dotenv -e .env.local -- npx prisma db pull",
     "prisma:generate": "prisma generate",
-    "prisma:migrate": "dotenv -e .env.local npx prisma migrate dev",
-    "prisma:studio": "dotenv -e .env.local npx prisma studio",
+    "prisma:migrate": "dotenv -e .env.local -- npx prisma migrate dev",
+    "prisma:migrate-create-only": "dotenv -e .env.local -- npx prisma migrate dev --create-only",
```
DafyddLlyr, line 19: Spent way too long trying to get the `--create-only` flag to be picked up here. 

Classic case of RTFM 🤦 - explainer in docs here why the additional `--` is required https://www.prisma.io/docs/orm/more/development-environment/environment-variables/managing-env-files-and-setting-variables#using-dotenv-cli-via-command-line

prisma/docs/how-to-update-prisma-schema.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+1. Update `schema.prisma`
+Make a change to this file which describes the change you would like to make (e.g. add column, add table).
+
+2. Generate migration
+Run `npm run prisma:migrate` - this will apply to the change to your local dev database, and generate a migration file in `/prisma/migrations`
+
+> [!TIP]
+> The command `npm run prisma:migrate-create-only` will generate a migration file without running it. This is helpful if you want to add any additional changes, and manually configure the migration. For example when renaming a column Prisma may drop and recreate (thus losing data). Manually generating an `ALTER table UPDATE column` SQL command allows us to work around this. You'll then need to run `npm run prisma:migrate` to apply your changes.
```
zz-hh-aa, line 13: Maybe I'm being silly but I don't quite understand where "Manually generating an `ALTER table UPDATE column` SQL command" should be happening

zz-hh-aa (approved):
Left one small q, super clear though!

prisma/docs/how-to-update-prisma-schema.md
DafyddLlyr: Thanks - updated the docs now. Is this any clearer?

Keen to jump on a call and demo this so we're all confident making changes 🙂

zz-hh-aa (approved):
I see! Thanks :)

PR #162, opened 2024-12-06
feat: Rename `build_prices` table, make fields non-nullable

DafyddLlyr:
## What does this PR do?
 - Changes table and column names to Postgres convention snake_case for the `buildprices` table
 - Updates code to reflect these fields as non-nullable

### DIFF
--- a/prisma/migrations/20241206160747_update_build_prices_names/migration.sql
+++ b/prisma/migrations/20241206160747_update_build_prices_names/migration.sql
@@ -1,0 +1,13 @@
+/*
+  Warnings:
+
+  - You are about to drop the `buildprices` table. If the table is not empty, all the data it contains will be lost.
+
+*/
+-- Alter table
+ALTER TABLE "buildprices" RENAME TO "build_prices";
+
+-- Alter table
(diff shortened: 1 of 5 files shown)


prisma/migrations/20241206160747_update_build_prices_names/migration.sql
```diff
@@ -0,0 +1,13 @@
+/*
+  Warnings:
+
+  - You are about to drop the `buildprices` table. If the table is not empty, all the data it contains will be lost.
```
zz-hh-aa, line 4: Should warnings be removed after the query is updated to use `RENAME` instead of `DROP`?

DafyddLlyr: Good spot thanks - yes! We should keep relevant warnings (very good for PRs) but remove redundant/incorrect ones.

DafyddLlyr: Please see ced8b008d76d8de52e3701e8633de730d6e7b107

zz-hh-aa (approved):
Thanks for tidying!

PR #164, opened 2024-12-06
feat: Rename `HPI` table columns, add not null constraints

DafyddLlyr:
## What does this PR do?
 - Renames `ladcode` to `lad_code`
 - Adds `NOT NULL` constraints to columns
 - Updates code to account for this constraint

### DIFF
--- a/app/data/hpiRepo.ts
+++ b/app/data/hpiRepo.ts
@@ -5,6 +5,7 @@ const getHPIByITL3 = async (itl3: string): Promise<number> => {
     const {
       _avg: { hpi2000: averageHpi },
     } = await prisma.hPI.aggregate({
+      // TODO: Should there be a relationship on ITL3?
       where: {
         itl3: {
           endsWith: itl3,
@@ -20,8 +21,7 @@ const getHPIByITL3 = async (itl3: string): Promise<number> => {
       throw new Error(`Data error: Unable to find hpi2000 for itl3 ${itl3}`);
     }
 
-
-    return averageHpi as number;
+    return averageHpi;
   } catch (error) {
     throw Error(`Data error: Unable to find hpi2000 for itl3 ${itl3}`);
   }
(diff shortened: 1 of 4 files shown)


app/data/hpiRepo.ts
```diff
@@ -5,6 +5,7 @@ const getHPIByITL3 = async (itl3: string): Promise<number> => {
     const {
       _avg: { hpi2000: averageHpi },
     } = await prisma.hPI.aggregate({
+      // TODO: Should there be a relationship on ITL3?
```
zz-hh-aa, line 8: Not quite sure what you mean by this?

prisma/migrations/20241206160747_update_build_prices_names/migration.sql
```diff
@@ -0,0 +1,13 @@
+/*
+  Warnings:
```
zz-hh-aa, line 2: Should we stick to the convention of removing warnings from the migrations?

prisma/migrations/20241206161734_make_build_prices_fields_required/migration.sql
```diff
@@ -0,0 +1,14 @@
+/*
+  Warnings:
```
zz-hh-aa, line 2: Same q as above!

app/data/hpiRepo.ts
DafyddLlyr: One to talk through next week and add to the backlog I think. We're currently just string matching on the start/end of an itl3 in multiple tables. 

What we might actually want to model is an ITL3 table, with relationships to other tables. This would be more efficient when it comes to queries and allow us to join multiple tables. Something like this maybe?

```ts
model Itl3 {
  id          Int       @id @default(autoincrement())
  code        String    @unique @db.VarChar(250)
  name        String    @db.VarChar(250)
  
  // Relationships
  GDHI        GDHI[]
  HPI         HPI[]
  Rent        Rent[]
  SocialRentEarnings SocialRentEarnings[]
  GasBills    GasBills[]
  ItlLookup   ItlLookup[]

  @@map("itl3")
}
```

zz-hh-aa: I _think_ I see what you mean! Will make a note and we can talk through it at next dev call :)

zz-hh-aa (approved):
Super, thanks!

PR #165, opened 2024-12-11
feat: update jest config for debugger

zz-hh-aa:
I wanted to be able to use the VSCode debugger on the tests and had to update our Jest configuration slightly to get it to work.

### DIFF
--- a/jest.config.ts
+++ b/jest.config.ts
@@ -84,16 +84,16 @@ const config: Config = {
   // ],
 
   // An array of file extensions your modules use
-  // moduleFileExtensions: [
-  //   "js",
-  //   "mjs",
-  //   "cjs",
-  //   "jsx",
-  //   "ts",
-  //   "tsx",
-  //   "json",
-  //   "node"
-  // ],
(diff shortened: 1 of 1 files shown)


gabrielegranello:
I think the principle makes sense to me. Just a quick question as I was not familiar with the syntax. I asked chatGPT what this lines do:
  transform: {                                                          
    "^.+\\.tsx?$": ["ts-jest", {
      tsconfig: "tsconfig.json"
    }]
  },    
  
  And it seems a fairly standard line to tell Jest to process any files ending in .ts or .tsx. However, how did it work before when transfer was set to none?

zz-hh-aa:
Good q, I did not understand myself so I asked Claude. 

Apparently, the line `preset: "ts-jest"` was sufficient for getting the tests to work; it automatically configures the Typescript transform. However to get VSCode debugger (especially if it's going to work with ts-jest) sometimes needs the configuration to be more explicit.

DafyddLlyr (approved):

PR #166, opened 2024-12-11
feat: migration rename columns in `gas_bills` table

zz-hh-aa:
# What does this PR do?
- Creates a new migration renaming some of the `gas_bills` columns
- Updates the schema accordingly

# Why?
I was starting on the data update migration and realised the shape of the gas bills data will change. It seemed like this would need to happen in two steps anyway (data change and schema change) so I figured we could rename the columns first, as the supplied CSVs are already in the next shape. 

# Questions
...do I need to update the tests to get everything passing even if we're just about to update the data anyway? 😅

### DIFF
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -91,8 +91,8 @@ model SocialRent {
 model GasBills {
   id     Int    @id
   region String @map("Region") @db.VarChar
-  itl    String @db.VarChar
-  bill   Float
+  itl1    String @db.VarChar
+  kwh_cost_pence   Float
 
   @@map("gas_bills")
 }
--- a/prisma/schema.prisma
+++ b/prisma/schema.prisma
@@ -90,7 +90,7 @@ model SocialRent {
 
 model GasBills {
   id     Int    @id
-  region String @map("Region") @db.VarChar
+  region String @map("region") @db.VarChar
   itl1    String @db.VarChar
   kwh_cost_pence   Float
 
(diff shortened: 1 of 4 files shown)


prisma/schema.prisma
```diff
@@ -91,8 +91,8 @@ model SocialRent {
 model GasBills {
   id     Int    @id
   region String @map("Region") @db.VarChar
-  itl    String @db.VarChar
-  bill   Float
+  itl1    String @db.VarChar
+  kwh_cost_pence   Float
```
DafyddLlyr, line 95: We also want to map this to camelCase 👍 

```suggestion
  kwhCostPence   Float @map("kwh_cost_pence")
```

DafyddLlyr (changes requested):
Once small comment!

prisma/schema.prisma
DafyddLlyr: Still missing an `@map` here which would mean we can use camelCase not snake_case in `getGasBillByITL3()`

DafyddLlyr (changes requested):

prisma/schema.prisma
zz-hh-aa: Ah thanks for catching! Missed that one, should be fixed now at [07d2f98](https://github.com/theopensystemslab/fairhold-dashboard/pull/166/commits/07d2f98c6ebbc685c87612e25af2d2df1091e494)

DafyddLlyr (approved):

PR #167, opened 2024-12-11
chore: tidy up migration docs and gitignore

zz-hh-aa:
# What does this PR do?
- Adds description of data loss warning to backup docs
- Changes gitignore to ignore whole backups directory (not just sql files)

# Why?
While the docs were super clear and I was able to carry out #166 smoothly, I knew when Daf's demo was less fresh in my mind I'd probably be very alarmed / confused by the data deletion warning, so figured it was best to clarify in the docs. 

I also backed up the db first, and found that it created a `.dump` file (rather than a `.sql` file which our gitignore expected) so just added the whole backups directory to the gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,4 +44,4 @@ next-env.d.ts
 .vscode
 
 # backups
-backups/*.sql
+backups/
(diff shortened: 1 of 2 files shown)


gabrielegranello (approved):
Thank you @zz-hh-aa  this is very valuable! Last week I wrote down a lot of notes as well after chatting with Daf as I was scared to forget the process too in a few weeks. Thanks for adding it to the documentation.

DafyddLlyr (approved):

PR #172, opened 2024-12-12
feat: rename `prices_paid` table

zz-hh-aa:
Changes table and column names to Postgres convention snake_case for the `prices_paid` table
Makes fields required

Basically follows #162!


gabrielegranello (approved):
Looks good to me 👍

PR #173, opened 2024-12-13
Url Search Parameters

gabrielegranello:
# What does this PR do?
It uses [URL Search Parameters](https://nextjs.org/docs/app/api-reference/functions/use-search-params) to partially fill the form. At the moment, it is set to work with the postcode.

# How does it work?
It uses the React hook `useSearchParams` that reads the parameters passed in the URL. The default value of the postcode in the form is set to be equal to the passed parameter.

# How to test it?
When the app is running on your `localhost`, for example `localhost:3000`, try to pass the postcode via the url.
For example, reload the page as `localhost:3000/?postcode=SE171PE`. You should find the postcode in the form to be equal to "SE171PE"


gabrielegranello:
Good point @DafyddLlyr . Now all the other parameters are available too. It can be tested with for example:
`http://localhost:3000/?postcode=SE171PE&houseSize=80&houseAge=1&houseBedrooms=2&houseType=T&maintenancePercentage=0.02`

PR #174, opened 2024-12-13
feat: make itl_lookup fields required

zz-hh-aa:
No renaming needed in this table. Makes the fields required.


app/data/itlRepo.ts
```diff
@@ -7,9 +7,6 @@ const getItl3ByPostcodeDistrict = async (
     const { itl3 } = await prisma.itlLookup.findFirstOrThrow({
       where: {
         postcode: postcodeDistrict,
-        itl3: {
-          not: null,
-        },
```
DafyddLlyr, line 12: Super nice how Prisma + TS catch these as being impossible conditions 👌

DafyddLlyr (approved):

PR #175, opened 2024-12-13
feat: rename `social_rent_adjustments` table and make fields required

### LINKED ISSUE #168: Add `NOT NULL` constraints to database where appropriate
Likely worth doing after https://github.com/orgs/theopensystemslab/projects/5/views/1?pane=issue&itemId=73543071

A number of our database columns lack a `NOT NULL` constraint (see the `?` after the column type in the Prisma schema). Infact, these columns always require a value and are always populated. Adding this constraint would more accurately model the data, and remove the need for redundant checks in our code.

zz-hh-aa:
Renamed what was `soc_rent_adjustements` to `social_rent_adjustments` (for consistency, didn't abbreviate the other one).

Also makes fields required.

I believe this is the last table that needs updating so when this merges it closes #168

### DIFF
--- a/prisma/migrations/20241213133842_rename_social_rent_adjustments/migration.sql
+++ b/prisma/migrations/20241213133842_rename_social_rent_adjustments/migration.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "soc_rent_adjustments" RENAME TO "social_rent_adjustments";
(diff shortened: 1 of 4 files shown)


DafyddLlyr (approved):

DafyddLlyr:
Really appreciate you picking up these tasks and working your way through all the tables @zz-hh-aa ! Codebase and data is in much better shape now 🙌

zz-hh-aa:
Cheers @DafyddLlyr, just following your example :) good tasks for learning!

PR #185, opened 2024-12-18
feat: rename gdhi column

zz-hh-aa:
While updating data I realised data naming pattern was inconsistent. I think we should stick to the pattern where there is no year suffix when the data is for the relevant year (2022) since that is what we are doing across all other tables, and only include a year suffix when the data is from a specific, older year (eg hpi_2000 because the social rent formula uses values from 2000).

Fixed (updated repo & renamed) version of #184

### DIFF
--- a/app/data/gdhiRepo.ts
+++ b/app/data/gdhiRepo.ts
@@ -10,7 +10,6 @@ const getGDHIByITL3 = async (itl3: string): Promise<number> => {
       },
       select: { gdhi: true },
     });
-    console.log('Result:', gdhi); 
     return gdhi;
   } catch (error) {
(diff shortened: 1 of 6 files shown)


app/data/gdhiRepo.ts
```diff
@@ -1,23 +1,23 @@
...
         AND: {
           itl3: { equals: itl3 },
         },
       },
-      select: { gdhi2020: true },
+      select: { gdhi: true },
     });
-
-    return gdhi2020;
+    console.log('Result:', gdhi);
```
DafyddLlyr, line 13: nit: Could we remove this?

DafyddLlyr (approved):

PR #186, opened 2024-12-18
feat: big data update to 2022 migration

zz-hh-aa:
I _think_ this is working a migration for the whole big data update! The various small migration PRs opened today emerged from this failing, and it now passes. 

A few small tasks in [`fairhold-data`](https://github.com/theopensystemslab/fairhold-data) have emerged which I'll work on next.


gabrielegranello (commented):
Sorry @zz-hh-aa if being slow here, but trying to remember the workflow.
This PR updates all the 2020 quantities e.g. gdhi, hpi and mean rent. Is it mean to add a column or replace it? The reason I am asking is because:
The gdhi table has 
`INSERT INTO "gdhi" ("id", "itl3", "region", "gdhi")`
the `gdhi` column 
The rent table has `monthly_mean_rent`
but he hpi table has `hpi_2000`.

Should all tables have a time stamp in the variable (_2020) or none of them, or is that a reason why only hpi needs one? Does this affect how we update the data every year or so?

zz-hh-aa:
Thanks for the eyes and q @gabrielegranello 

The convention I was adhering to was that most of the data should not specify a year (because it is all for the most recent year for which all of the data is available, now 2022) _except_ the HPI data. This is because HPI data is for 1999 (the 2000 data is pre-update) because this is what the social rent formula uses as an input. 

In sum, no year specified = 2022. 

Does this make sense as a convention to follow and, if so, any thoughts on the best place to explain it?

gabrielegranello:
Thanks for explaining @zz-hh-aa , that makes sense to me. Not sure what is the best practice here probably @DafyddLlyr can offer better advice but I was wondering if we should we add something in the readme file? Something like 
-- All data without a specified year is for the most recent complete year 
-- HPI data explicitly references 1999 due to its relevance to the social rent formula.
or similar.

Or should we add a specific readme to the [Postgres database](https://vercel.com/open-systems-labs-projects/fairhold-dashboard/stores/postgres/store_glZ9OcBmnsANxljr/data) itself where we describe the data?

DafyddLlyr:
@zz-hh-aa Could you please rebase this PR to `main`?

I'm getting the following issues when testing this PR locally - 

```sh
➜  fairhold-dashboard git:(oz/update-data-to-2022) pnpm prisma:migrate

> fairhold-dashboard@0.1.0 prisma:migrate /Users/dafyddllyr/Code/fairhold-dashboard
> dotenv -e .env.local -- npx prisma migrate dev

Prisma schema loaded from prisma/schema.prisma
Datasource "db": PostgreSQL database "fairhold", schema "public" at "localhost:5400"

- Drift detected: Your database schema is not in sync with your migration history.

The following is a summary of the differences between the expected database schema given your migrations files, and the actual schema of the database.

It should be understood as the set of changes to get from the expected schema to the actual schema.

[*] Changed the `hpi` table
  [-] Removed column `hpi_2000`
  [+] Added column `hpi_1999`

[*] Changed the `itl_lookup` table
  [-] Removed column `district`
  [+] Added column `region`

[*] Changed the `prices_paid` table
  [-] Removed column `transaction_identifier`

- The migrations recorded in the database diverge from the local migrations directory. Last common migration: `20241218105236_rename_gdhi_column`. Migrations applied to the database but absent from the migrations directory are: 20241218155911_update_hpi_year, 20241218162252_remove_transaction_id_column, 20241218161553_rename_district_to_region
```

It might not be related to this, but if you're please able to rebase we can at least rule it out 👌

zz-hh-aa:
Cheers @DafyddLlyr that's rebased now, double checked that tests were succeeding locally for me too post-rebase!

DafyddLlyr:
Thanks @zz-hh-aa !

I'm now hitting this error locally - 

```sh
Applying migration `20241218111832_update_data_to_2022`
Error: P3018

A migration failed to apply. New migrations cannot be applied before the error is recovered from. Read more about how to resolve migration issues in a production database: https://pris.ly/d/migrate-resolve

Migration name: 20241218111832_update_data_to_2022

Database error code: 42703

Database error:
ERROR: column "hpi_2000" of relation "hpi" does not exist

Position:
138 (176, 'TLK43', 'Devon CC', '22209')
139 ;
140
141 -- Update hpi table
142 TRUNCATE TABLE "hpi";
143 INSERT INTO "hpi" ("id", "itl3", "region", "hpi_2000") 

DbError { severity: "ERROR", parsed_severity: Some(Error), code: SqlState(E42703), message: "column \"hpi_2000\" of relation \"hpi\" does not exist", detail: None, hint: None, position: Some(Original(6144)), where_: None, schema: None, table: None, column: None, datatype: None, constraint: None, file: Some("parse_target.c"), line: Some(1065), routine: Some("checkInsertTargets") }
```

This is happening because the column is called `hpi_1999` currently, however, we're trying to insert into `hpi_2000` - 

https://github.com/theopensystemslab/fairhold-dashboard/blob/oz/update-data-to-2022/prisma/schema.prisma#L34

zz-hh-aa:
Interesting, I wonder why it works locally for me--does your local branch have the HPI year migration after the data update migration?
image

DafyddLlyr:
@zz-hh-aa Good point - the order of migrations is correct for me also (data added before column renamed).

I think the issue comes from the fact we’re slightly mixing up migrations (db structure) and seeding (db content) - it might be best to keep a cleaner separation of the two.

Here's what I'd propose - 
 - We don't make this a migration
 - We just run the raw SQL to update the production DB (we can also run it locally to test)

So I think that we should end up with a .sql file (in the fairhold-data repo) which we can locally run and test, check it's ok, then manually run on production?

One other reason for this is once we set up foreign keys between tables, the truncate + insert method won't work in future years as expected.

PR #187, opened 2024-12-18
feat: update column name to hpi1999

zz-hh-aa:
The data update changes HPI values from 2000 to 1999, the correct input for social rent calculations.

This migration makes that change in the schema and relevant repo / service.

### DIFF
--- a/app/data/hpiRepo.test.ts
+++ b/app/data/hpiRepo.test.ts
@@ -1,5 +1,5 @@
 // __tests__/hpiRepo.test.ts
-import { hpi2000Repo } from "./hpiRepo"; // Adjust the import according to your file structure
+import { hpi1999Repo } from "./hpiRepo"; // Adjust the import according to your file structure
 import prisma from "./db"; // Your Prisma setup file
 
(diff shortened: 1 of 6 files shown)


app/data/hpiRepo.test.ts
```diff
@@ -1,5 +1,5 @@
 // __tests__/hpiRepo.test.ts
-import { hpi2000Repo } from "./hpiRepo"; // Adjust the import according to your file structure
+import { hpi1999Repo } from "./hpiRepo"; // Adjust the import according to your file structure
 import prisma from "./db"; // Your Prisma setup file
```
DafyddLlyr, line 3: nit: These comments look like leftover AI fragments 🗑️ 🧹

DafyddLlyr (approved):

app/data/hpiRepo.test.ts
zz-hh-aa: Heh I caught them but then thought removing them in this PR might be bad practice, have now cleaned them up 🧹

PR #211, opened 2025-01-06
chore: tsconfig linting

zz-hh-aa:
~~Not sure if anyone else had this lingering change locally (I believe it happened after a dependabt eslint update) but `tsconfi.json` was updated and I didn't want to lump that commit into another PR. Here's a separate PR for the minor changes!~~

Ah-ha, I think we have a conflict between what eslint wants the `tsconfig.json` to look lkike and what `dev:db` wants it to look like. Let's discuss later!

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,10 @@
 {
   "compilerOptions": {
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": [
+      "dom",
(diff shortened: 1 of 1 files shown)


DafyddLlyr (approved):
Yep running `npm lint` gives me the same result here - maybe there was a version bump somewhere via depenedabot which necessitated this change.

Issue #202, opened 2025-01-06
Fix new Next error


DafyddLlyr:
"A component is changing an uncontrolled input to be controlled. This is likely caused by the value changing from undefined to a defined value, which should not happen. Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://react.dev/link/controlled-components"

Ignored frames:
``` createUnhandledError
..\src\client\components\react-dev-overlay\internal\helpers\console-error.ts (14:35)
formattedErrorMessage
..\src\client\components\react-dev-overlay\internal\helpers\use-error-handler.ts (30:34)
console.error
..\src\client\components\globals\intercept-console-error.ts (32:11)
updateProperties
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (17835:1)
commitUpdate
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (19083:1)
runWithFiberInDEV
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (543:1)
commitHostUpdate
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11142:1)
commitMutationEffectsOnFiber
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (12185:1)
recursivelyTraverseMutationEffects
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11931:1)
commitMutationEffectsOnFiber
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11942:1)
recursivelyTraverseMutationEffects
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11931:1)
commitMutationEffectsOnFiber
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11942:1)
recursivelyTraverseMutationEffects
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11931:1)
commitMutationEffectsOnFiber
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11942:1)
recursivelyTraverseMutationEffects
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11931:1)
commitMutationEffectsOnFiber
node_modules\next\dist\compiled\react-dom\cjs\react-dom-client.development.js (11942:1)
recursivelyTraverseMutationEffects
[...]

PR #216, opened 2025-01-06
Fix new Next error (task #202)

gabrielegranello:
# What does this PR do?

It fixes the error #202 .

# How?
It sets the value of the Input component to an empty string should this not be set by the url optional parameters


DafyddLlyr (approved):
Perfect!

PR #224, opened 2025-01-07
Broken labels fix

### LINKED ISSUE #217: bug: Broken labels for maintenance
https://github.com/user-attachments/assets/5e67544f-642f-4202-8932-8436ec1892c8

It looks like we have mismatched or repeated label names here

gabrielegranello:
# What does this PR do?
It fixes #217 .

# How?
It assigns unique identifiers to the radio group options .

https://github.com/user-attachments/assets/d9ceb825-8a4c-4ab9-9f31-4c3184585a2b

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -162,21 +162,21 @@ const CalculatorInput = () => {
                       onValueChange={(value) => field.onChange(value)} // Bind selection to field
                     >
                       <div className="flex items-center space-x-2">
-                        <RadioGroupItem value="D" id="option-one" />
-                        <Label htmlFor="option-one">Detached</Label>
+                        <RadioGroupItem value="D" id="radio_option_D" />
+                        <Label htmlFor="radio_option_D">Detached</Label>
                       </div>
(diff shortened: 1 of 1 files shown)


app/components/ui/CalculatorInput.tsx
```diff
@@ -162,21 +162,21 @@ const CalculatorInput = () => {
                       onValueChange={(value) => field.onChange(value)} // Bind selection to field
                     >
                       <div className="flex items-center space-x-2">
-                        <RadioGroupItem value="D" id="option-one" />
-                        <Label htmlFor="option-one">Detached</Label>
+                        <RadioGroupItem value="D" id="radio_option_D" />
```
DafyddLlyr, line 165:
```suggestion
                        <RadioGroupItem value="D" id="radio-option-D" />
```

We should stick to kebab-case for these variables 👍

DafyddLlyr (approved):
One small tidy up required - thanks for the quick fix here!

app/components/ui/CalculatorInput.tsx
gabrielegranello: Thanks for the quick review!

PR #228, opened 2025-01-07
feat: Create `prices_paid_summary` table

DafyddLlyr:
## What does this PR do?
 - Adds a new `PricesPaidSummary` model - this will hold pre-calculated values emulating the business logic used in `pricesPaidRepo.rs`
 - The data will be populated via this script once merged to production https://github.com/theopensystemslab/fairhold-data/pull/1
 - The model is implemented here - https://github.com/theopensystemslab/fairhold-dashboard/pull/241
   - This needs to be a separate, follow-on, PR as it will rely on the table being already populated


zz-hh-aa (approved):
Very cool / clever way of speeding up our query, hadn't encountered `UNIQUE INDEX` before!

PR #229, opened 2025-01-07
fix: Ensure input postcode is always parsed

DafyddLlyr:
## What's the problem?
Whilst working on https://github.com/theopensystemslab/fairhold-dashboard/pull/228 I noticed that postcode district, area, and sector were `undefined`.

This is due to a regression introduced here - https://github.com/theopensystemslab/fairhold-dashboard/pull/65

Without using the `calculationSchema` the input postcode isn't parsed into its constituent parts, allowing queries to fail. 

This happens due to this behaviour - https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/null-and-undefined#null-and-undefined-in-queries-that-affect-many-records

The returned itl3 is always the same one, `TLH23`.

Also - the frontend is not correctly validating postcodes, it's currently possible to pass an invalid postcode to the API.

## What's the solution?
Revert change to always parse input via the calculation schema.

image


app/api/route.ts
```diff
@@ -8,12 +7,7 @@ export async function POST(req: Request) {
   try {
     // Parse and validate user input
     const data = await req.json();
-    let input: api;
-    if (!apiSchema.safeParse(data).success) {
-      input = calculationSchema.parse(data);
-    } else {
-      input = data;
-    }
+    const input = calculationSchema.parse(data);
```
DafyddLlyr, line 10: Main change in this PR - falling back to parsing the input.

app/schemas/formSchema.ts
```diff
@@ -11,17 +11,17 @@ const HouseTypeEnum = z.enum(HOUSE_TYPES);
...
+  houseAge: z.coerce.number().positive("House age must be a positive number"),
   houseBedrooms: z.coerce
     .number()
-    .positive("houseBedrooms must be a positive integer"),
+    .positive("House bedrooms must be a positive number"),
   houseType: HouseTypeEnum.refine(
     (value) => HouseTypeEnum.options.includes(value),
     {
-      message: `houseType is required and must be one of ${HOUSE_TYPES}`,
+      message: `House type is required`,
```
DafyddLlyr, line 24: Error messages here are using facing and more a bit more user-friendly.

gabrielegranello (commented):
Thanks @DafyddLlyr !
Can I quickly double check if I get the logic right here?
We have two schemas: `calculationSchema` and `formSchema`.
 The main difference is that:
 1) formSchema checks if the the string is a valid postcode with the following:
```
housePostcode: z
    .string()
    .min(1, "Postcode is required")
    .refine(isValidPostcode, "Invalid postcode"),
```
 We are using this for form validation. 
( side note, it is interesting to me that we can check if a postcode is valid without transforming the string into a postcode object, great thing)

2) `calculationSchema`:
it checks the string to be a valid postcode AND it transforms it into a postcode object
 ```
housePostcode: z
    .string()
    .min(1, "housePostcode is required")
    .refine(isValidPostcode, "Invalid postcode")
    .transform(parsePostcode)
    .refine((postcode): postcode is ValidPostcode => postcode.valid), 
```

we are using this schema to ALWAYS parse the data before sending it to the API. Meaning to the API we are passing always a postcode object.

Question:  
what does this extra final line do?
`.refine((postcode): postcode is ValidPostcode => postcode.valid), `

Is it like a double check that after transforming a valid string into an object, it is still a valid postcode?

**Comment by gabrielegranello** [comment]
Oh side note. I can see a third schema called `apiSchema` in the folder. I don't think we are using it in route.ts so I tried deleting it. Is it perhaps an old file that stuck around?

I am scary to say the words ' everything seems working without it', bit it actually does

**Comment by DafyddLlyr** [comment]
@gabrielegranello Yep that's correct - we always want to parse in the backend as it could technically be accessed from anywhere. We also need the postcode object here so we're doing this transformation as part of the schema.

This line is a type predicate ([docs](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)) which narrows down the type for us (the library isn't doing this automatically).

```ts
.refine((postcode): postcode is ValidPostcode => postcode.valid), 
```

gabrielegranello (approved):

PR #230, opened 2025-01-07
dp/link prisma versions

DafyddLlyr:
- **chore: Upgrade Prisma to 6.2.0**
- **chore: Group prisma dependencies**


gabrielegranello (approved):
Thanks for checking this and updating the yml file!

PR #231, opened 2025-01-07
UI - Input layout

gabrielegranello:
# What does this PR do?
This is a work in progress on the UI of the calculator. It sets the font style and the initial grid. 
It also adds a further component called `FormEdit.tsx` in the dashboard, so that the user knows which input data have been used for the analysis.

Webflow to match here https://fairhold.webflow.io/classbook


app/components/ui/CalculatorInput.tsx
```diff
@@ -25,6 +25,7 @@ import { Label } from "@/components/ui/label";
 const CalculatorInput = () => {
   const [view, setView] = useState("form");
   const [data, setData] = useState<Household | null>(null);
+  const [formData, setFormData] = useState<formType | null>(null);
```
DafyddLlyr, line 28: I don't think we need to hold these values in state a second time - we can already access this with `methods.getValues()`

I'm not sure if `methods` is a standard naming convention - maybe we just call it `form`?

app/components/ui/CalculatorInput.tsx
```diff
@@ -234,7 +333,12 @@ const CalculatorInput = () => {
       </div>
     );
   } else if (view === "dashboard") {
-    return <Dashboard data={data as Household} />;
+    return (
+      <Dashboard
+        processedData={data as Household}
+        inputData={formData as formType}
```
DafyddLlyr, line 339: See previous comment - I think the below should work.

```suggestion
        inputData={methods.getValues()}
```

app/globals.css
DafyddLlyr, line 1: I hadn't appreciated how much of this was redundant and just webflow styles (`w-*`).

Maybe it would be simpler to not import this at all and just emulate / copy styles ad-hoc. What do you think?

DafyddLlyr: Actually - here's a better idea 💡 

If we just make a file called `webflow.css` which is just copy/pasted from the source, and then we can use `import "./webflow.css"` in `layout.tsx`

This should give us easy access to the styles, without cluttering our styles?

gabrielegranello:
Sorry almost good to go, keep failing the coverage test. I am going to try a couple of other times, otherwise I night just add a couple of extra lines in the test to increase the coverage

app/components/ui/Dashboard.tsx
```diff
@@ -1,12 +1,14 @@
 import React, { useRef, useState } from "react";
 import GraphCard from "./GraphCard";
 import { Household } from "@/app/models/Household";
+import { formType } from "@/app/schemas/formSchema";
 
 interface DashboardProps {
-  data: Household;
+  processedData: Household;
+  inputData: formType; // Add this property
```
DafyddLlyr, line 8: Leftover GPT comment? 🤖

app/components/ui/Dashboard.tsx
```diff
@@ -1,12 +1,14 @@
 import React, { useRef, useState } from "react";
 import GraphCard from "./GraphCard";
 import { Household } from "@/app/models/Household";
+import { formType } from "@/app/schemas/formSchema";
```
DafyddLlyr, line 4:
```suggestion
import type { Form } from "@/app/schemas/formSchema";
```

Types and interfaces should be in PascalCase, and if we can avoid it we shouldn't use `Type` suffixes.

app/components/ui/FormEdit.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  if (houseType === "T") {
+    return "Terrace";
+  }
+
+  if (houseType === "F") {
+    return "Flat";
+  }
+}
```
DafyddLlyr, line 32:
```suggestion
const houseTypeDisplayNames: Record<HouseType, string> = {
  D: "Detached",
  S: "Semi-detached",
  T: "Terrace",
  F: "Flat",
};
```

And then later - `houseTypeDisplayNames[formData.houseType]`

Using a plain object for a map instead of a function is probably simpler, more maintainable, and more type safe (adding a new house type will throw an error here, as will missing one out).

app/components/ui/Dashboard.tsx
```diff
@@ -1,12 +1,14 @@
...
 
 interface DashboardProps {
-  data: Household;
+  processedData: Household;
+  inputData: formType; // Add this property
 }
 
-const Dashboard: React.FC<DashboardProps> = ({ data }) => {
+const Dashboard: React.FC<DashboardProps> = ({ inputData }) => {
```
DafyddLlyr, line 11:
```suggestion
const Dashboard: React.FC<DashboardProps> = () => {
```

This would avoid needing the work around for an unused variable.

app/components/ui/CalculatorInput.tsx
```diff
@@ -36,7 +36,7 @@ const CalculatorInput = () => {
   const urlHouseType = searchParams.get("houseType");
   const urlMaintenancePercentage = searchParams.get("maintenancePercentage");
 
-  const methods = useForm<formType>({
+  const form = useForm<formType>({
```
DafyddLlyr, line 39: Much better name 👍

app/components/ui/CalculatorInput.tsx
```diff
@@ -239,10 +336,15 @@ const CalculatorInput = () => {
...
-  if (view === "dashboard" && data) {
-    return <Dashboard data={data} />;
+
+  } else if (view === "dashboard") {
+    return (
+      <Dashboard
+        processedData={data as Household}
+        inputData={form.getValues()}
+      />
+    );
```
DafyddLlyr, line 346: It looks like the changes made in #243 have been lost when dealing with the merge conflicts here.

DafyddLlyr (changes requested):
Visually great 👍 

A few small comments and suggestions on the code.

app/components/ui/CalculatorInput.tsx
gabrielegranello: Thanks @DafyddLlyr I should have picked up what was lost when merging

app/components/ui/CalculatorInput.tsx
```diff
@@ -74,175 +74,276 @@ const CalculatorInput = () => {
...
-  if (view === "dashboard" && data) {
-    return <Dashboard data={data} />;
+  } else if (view === "dashboard" && data) {
+    const formValues = form.getValues();
+    return (
+      <Dashboard
+        processedData={data}
+        inputData={formSchema.parse(formValues)}
+      />
+    );
```
DafyddLlyr, line 346:
```suggestion
  }
  
  if (view === "dashboard" && data) {
    const formValues = form.getValues();
    return (
      <Dashboard
        processedData={data}
        inputData={formValues}
      />
    );
```

At this point, the values are already parsed so we don't need to repeat this 👍

DafyddLlyr (changes requested):
One small comment on code 👍 

Also the inputs' text colours need to be black / darker than the placeholder colour -

app/components/ui/CalculatorInput.tsx
gabrielegranello: Hopefully fixed thanks @DafyddLlyr

DafyddLlyr (approved):
Great!

PR #238, opened 2025-01-07
chore: Enable `strictUndefinedChecks` flag on Prisma

DafyddLlyr:
Follow up to https://github.com/theopensystemslab/fairhold-dashboard/pull/229 - this should have helped us avoid this issue in the first instance.

Docs: https://www.prisma.io/docs/orm/prisma-client/special-fields-and-types/null-and-undefined

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,6 +9,7 @@
     "skipLibCheck": true,
     "strict": true,
     "noEmit": true,
+    "exactOptionalPropertyTypes": true,
     "esModuleInterop": true,
     "module": "esnext",
     "moduleResolution": "bundler",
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -9,6 +9,7 @@
     "skipLibCheck": true,
     "strict": true,
     "noEmit": true,
+    "exactOptionalPropertyTypes": true,
```
zz-hh-aa, line 12: I almost asked about this line but think I get it now through Prisma docs:

> While `strictUndefinedChecks` will raise runtime errors for invalid undefined usage, `exactOptionalPropertyTypes` will catch these issues during the build process.

Cool!

zz-hh-aa (approved):
Great!

PR #241, opened 2025-01-08
feat: Query new `PricesPaidSummary` model

DafyddLlyr:
## What does this PR do?
 - Follows on from #228
 - Replaces the business logic with a simple query to the pre-computed table
 - This can't actually be used or tested until #228 is merged and the data is populated on production

### DIFF
--- a/app/api/route.test.ts
+++ b/app/api/route.test.ts
@@ -82,7 +82,9 @@ describe("POST API Route", () => {
 
     // Assertions for the expected error response
     expect(NextResponse.json).toHaveBeenCalledWith(
-      expect.objectContaining({ error: expect.any(String) }),
+      expect.objectContaining({
+        error: expect.objectContaining({ code: "UNHANDLED_EXCEPTION" }),
+      }),
       { status: 500 }
     );
   });
@@ -112,7 +114,7 @@ describe("POST API Route", () => {
     expect(calculationService.getHouseholdData).toHaveBeenCalledWith({
       ...validApiInput,
       // Parsed postcode object
-      housePostcode: { 
+      housePostcode: {
         area: "SE",
         district: "SE17",
         incode: "1PE",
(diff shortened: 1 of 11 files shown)


zz-hh-aa:
Do we need to run the migration from `fairhold-data` [PR 1](https://github.com/theopensystemslab/fairhold-data/pull/1) over here to test this properly? It looks like the `pricesPaidSummary` table has been created but not populated. 

After running `backup`, `restore` etc., I'm now getting the new data through locally (checked in Prisma Studio) but `pricesPaidSummary` is empty so I'm hitting errors when trying to test the new query.

DafyddLlyr:
This is now running against the populated `prices_paid_summary` table on Vercel.

The difference isn't actually as much as I'd hoped when comparing with prod - it's notably faster running locally. I still think this approach is good but it might need a little more work.

gabrielegranello:
Oh I was trying to test it locally but I fail the call to the database.
After running backup and restore, I can see the correct database ( in fact I can see the pricesPaidSummary table. Am I missing something to properly try and test this code locally?

DafyddLlyr:
@gabrielegranello Have you ran `npm run prisma:generate` to update your local client for this branch?

BTW, this runs as a post-install script on `npm i` so it's often just easier/quicker to run `npm i` 👌

gabrielegranello:
Oh I actually did not. After running it's working but can I ask why? My understanding was that prisma:generate should be run when the local schema of the data changes i.e. if we touch `prisma.schema`. When I see the files changed in this PR they are all type `repo` or type `test`

DafyddLlyr:
If that resolved the issue, I would guess that you've not ran it since https://github.com/theopensystemslab/fairhold-dashboard/pull/228 merged which included a new Prisma model which this PR relies on.

gabrielegranello:
Oh got it, schema changed with #228 I think

gabrielegranello (approved):
Very tidy! Sorry it took me a while to understand what was going on :)  My understanding is:
1) With another PR, you created a new table called `pricesPaidSummary`, where the data are already aggregated by postcode.
2) When querying to such a table, you query simultaneously  for all three levels of granularity.
3) The results of the query will be actually given in the right order (sector, district, area)
4) The first one is retained

PR #242, opened 2025-01-08
UI Scroll layout for the graphs

gabrielegranello:
# What does this PR do?

It sets a series of scrollable layouts for where the graphs are going to live.
The scrolling occurs between `GraphCard1` and `GraphCard6`, which will show the different graphs and text.

https://github.com/user-attachments/assets/2af7c80c-b928-43e0-94f0-8dbf4c8185d9


app/components/ui/Dashboard.tsx
```diff
@@ -1,27 +1,46 @@
...
+import GraphCard6 from "./GraphCard6";
 import { Household } from "@/app/models/Household";
 
 interface DashboardProps {
   data: Household;
 }
 
 const Dashboard: React.FC<DashboardProps> = ({ data }) => {
+  const handleNext = () => {
+    const container = document.querySelector(".snap-scroll") as HTMLElement;
```
DafyddLlyr, line 16: We should be aiming to use `ref` in React instead of calling native DOM functionality like this.

DafyddLlyr (commented):
Video looks great! 👏 

In terms of implementation, if we want to make a change to layout or styling we now need to make it in 6 places and keep things in parallel - this is super hard and an easy way to introduce mistakes and inconsistencies.

I'd recommend having a crack at pattern like this - 
- A single `GraphCard` component
- It accepts a title prop
- It accepts `children` (ie the "content" / graphs - have a look at `React.PropsWithChildren`)
- It's responsible for layout, next button, scroll, title etc - but not the actual content

app/components/ui/Dashboard.tsx
gabrielegranello: Does this version look better? Sorry I wasn't familiar with the `children` concept , it seems pretty powerful.
If I understand it correctly, it means I can pass other optional React components to the card. For example, I could pass a title AND a graph, or a title AND a section. So the basically the GraphCard component should provide the 'scrollability feature' and render the content that is passed to it, is that correct?

app/components/ui/Dashboard.tsx
```diff
@@ -1,27 +1,52 @@
...
+        behavior: "smooth",
+      });
+    }
+  };
+
+  // Dummy identity function to avoid linting errors
+  const identity = (data: Household) => {
+    return data;
+  };
+  identity(data);
```
DafyddLlyr, line 24: I don't really follow what this is for sorry!

app/components/ui/Dashboard.tsx
DafyddLlyr: Perfect! Super quick turnaround 🏅

DafyddLlyr (approved):
One small question about `identity()` - but this is great!

app/components/ui/Dashboard.tsx
gabrielegranello: Oh sorry, it was a trick to avoid listing errors. We are passing `data` to the `Dashboard` component. That makes sense because `data` will be necessary to generate the plots in the future.
The problem is that right now we are doing nothing with it, so it flags an error of a variable not being used. So I created a function that returns itself for the moment to get around it. I considered passing a plot to one of the Card, but I figured it would be better to be a separate PR

app/components/ui/Dashboard.tsx
```diff
@@ -1,27 +1,52 @@
...
+        <GraphCard title="What would you choose?"></GraphCard>
+      </div>
+
+      <div className="fixed bottom-4 right-4">
+        <button
+          onClick={handleNext}
+          className="px-4 py-2 bg-blue-500 text-white rounded hover:bg-blue-600"
+        >
+          Next
+        </button>
```
DafyddLlyr, line 48: Just realised - this currently remains visible on the final card.

A few ways we could handle this - a quick/simple version might be to just to track the index of the pages on `handleNext()`, and hide for the final one.

DafyddLlyr (changes requested):
Sorry just noticed one small thing!

app/components/ui/Dashboard.tsx
DafyddLlyr: Ah right! That makes sense - thanks for the explanation 👍 

There's an easy option here! We can add `// eslint-disable-next-line <RULE_NAME>` as a comment to the line above an unused variable we want to keep and it should resolve this issue.

```ts
// eslint-disable-next-line @typescript-eslint/no-unused-vars
```

Docs: https://eslint.org/docs/latest/use/configure/rules#disabling-rules

app/components/ui/Dashboard.tsx
gabrielegranello: Thanks for the suggestion! Hopefully this will do

gabrielegranello: Uploading Screen Recording 2025-01-09 at 09.57.41.mov…

PR #244, opened 2025-01-08
feat: Gracefully handle non-English postcodes

DafyddLlyr:
## What does this PR do?
 - Adds a custom `APIError` class
 - Implements this where we have specific errors which require feedback for the user
 - Handles these errors on the frontend and updates the form errors accordingly
 - Closes https://github.com/theopensystemslab/fairhold-dashboard/issues/232

https://github.com/user-attachments/assets/4722a645-afdd-42d5-82fd-aeb03a13de14

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -21,6 +21,7 @@ import {
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
 import { Label } from "@/components/ui/label";
+import { APIError } from "@/app/lib/exceptions";
 
 type View = "form" | "loading" | "dashboard";
 
@@ -68,9 +69,30 @@ const CalculatorInput = () => {
       body: JSON.stringify(data),
     });
     const processedData = await response.json();
+    if (!response.ok) return handleErrors(processedData.error);
+
     setData(processedData);
     setView("dashboard");
   };
+
+  const handleErrors = (error: APIError) => {
+    switch (error.code) {
+      case "ITL3_NOT_FOUND":
+      case "INSUFFICIENT_PRICES_PAID_DATA":
+        methods.setError(
+          "housePostcode", 
+          { message: 
+            "Insufficient data for this postcode. Please try again with a different postcode" 
+          }
+        );
+        break;
+      case "UNHANDLED_EXCEPTION":
+      default:
+        console.error(error)
+    };
+    
+    setView("form");
+  }
 
   if (view === "form") {
     return (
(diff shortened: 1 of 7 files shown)


app/components/ui/CalculatorInput.tsx
```diff
@@ -64,10 +65,31 @@ const CalculatorInput = () => {
...
+    switch (error.code) {
+      case "ITL3_NOT_FOUND":
+      case "INSUFFICIENT_PRICES_PAID_DATA":
+        methods.setError(
+          "housePostcode", 
+          { message: 
+            "Insufficient data for this postcode. Please try again with a different postcode" 
+          }
+        );
+        break;
```
DafyddLlyr, line 84: We could handle these two cases differently, but right now this is a little white-lie about non-English postcodes.

app/data/itlRepo.ts
```diff
@@ -15,10 +16,11 @@ const getItl3ByPostcodeDistrict = async (
 
     return itl3;
   } catch (error) {
-    throw new Error(
-      `Data error: Unable get get itl3 for postcode district ${postcodeDistrict}`
-    );
-
+    throw new APIError({
+      status: 500,
+      message: `Data error: Unable get get itl3 for postcode district ${postcodeDistrict}`,
```
gabrielegranello, line 21: minor typo 'get get'

DafyddLlyr: Great catch - fixed in [ecd1570](https://github.com/theopensystemslab/fairhold-dashboard/pull/244/commits/ecd157052766cb0aeb602f60ec69f8463b3c85ce)

gabrielegranello (approved):
That's very elegant! If I get it correctly:
1) A new object `APIErrorCode` is created. This extends the current `Error` class
2) `APIErrorCode` can be of 3 types:  `| "ITL3_NOT_FOUND" | "INSUFFICIENT_PRICES_PAID_DATA" | "UNHANDLED_EXCEPTION"`;
3) If the Error code type 1 or 2 is returned by the API, we trigger the form validation with message "Insufficient data for this postcode. Please try again with a different postcode" 
Otherwise we trigger `console.log ` with the default error message from the general `Error` class. I assume errors such as : failing to connect to the database or similar will follow in this category, is that correct?

DafyddLlyr:
@gabrielegranello Perfectly correct! 

In future it would be nice to handle unhandled erorrs a bit better (e.g. logging the error to an eternal monitoring service, displaying a "try again" status to the user), but for now we just return the user the specific errors we care most about and throw unhandled errors.

PR #256, opened 2025-01-09
feat: calculate gas bills using new `kwh_cost_pence` figures

### LINKED ISSUE #112: Make code changes following data update
- [ ] using pence per kwh instead of house average, so can remove some bills math

Blocked by [data update](https://github.com/theopensystemslab/fairhold-dashboard/issues/110)

zz-hh-aa:
A tidied up version of #195 (info copied below for reference)

Closes #112 and #26

---
>What does this PR do?
Adds new properties gasBillExistingBuildYearly and gasBillNewBuildOrRetrofit and methods to calculate them to Household class
Creates new objects to store kwh cost in pence in constants: KWH_M2_YR_EXISTING_BUILDS, KWH_M2_YR_NEWBUILDS_RETROFIT
Calculates inflating gas bills in Lifetime
Why?
There is no test associated with the new gas bill figures yet because the Lifetime.test.ts file doesn't exist yet (exists in another open PR). So this will need to be rebased after oz/depreciation-update (and chain) get merged!

Additionally, since the table no longer shows gas bills but gas prices, I have renamed the table / column / repo / service / tests accordingly in a migration. `Household` still shows `gasBill...` because there the `kwh_cost_pence` is calculated into the bill cost.


app/models/constants.ts
```diff
@@ -146,4 +146,25 @@ export const HOUSE_BREAKDOWN_PERCENTAGES: houseBreakdownType = {
     depreciationPercentageYearly: .0343,
     percentOfMaintenanceYearly: .056
   },
+}
+export type billsByHouseType = { // TODO: fix type syntax!
+  F: number;
+  T: number;
+  S: number;
+  D: number;
+}
```
DafyddLlyr, line 155:
```suggestion
}

export type BillsByHouseType = Record<HouseType, number>
```

app/models/constants.ts
```diff
@@ -146,4 +146,25 @@ export const HOUSE_BREAKDOWN_PERCENTAGES: houseBreakdownType = {
...
+
+/* Figures are from LETI / Studio PDP */
+export const KWH_M2_YR_EXISTING_BUILDS = {
```
DafyddLlyr, line 158:
```suggestion
export const KWH_M2_YR_EXISTING_BUILDS: BillsByHouseType = {
```

app/models/constants.ts
DafyddLlyr, line 1: I think these are the type fixes you're looking for! 😄

DafyddLlyr (approved):

PR #257, opened 2025-01-09
feat: Initial drawer setup

DafyddLlyr:
## What does this PR do?
 - Adds a `Drawer` component based on the shadcn/ui [Sheet](https://ui.shadcn.com/docs/components/sheet)
 - Creates an `MaintenanceExplainerDrawer` which uses this drawer, with the required (initial) content
 - This pattern can be easily expanded upon in graphs and for other form inputs as required

Hit a few issues here with colours and our mix of tailwind and CSS variables - we should probably stick to one (Tailwind?) but we can discuss this tomorrow and fix forward.

https://github.com/user-attachments/assets/76230a05-6f9f-4e9d-b0f4-67c1b47ede13

### DIFF
--- a/app/components/ui/MaintenanceExplainerDrawer.tsx
+++ b/app/components/ui/MaintenanceExplainerDrawer.tsx
@@ -1,0 +1,36 @@
+import { Drawer } from "./Drawer";
+
+export const MaintenanceExplainerDrawer = () => (
+  <Drawer
+    title="How much will you spend on maintenance and improvements?"
+    buttonTitle="What is this?"
+    description="Fairhold treats homes as durables: their value depends on whether they are well cared-for. If you own a home you probably already spend more than you think on occasional maintenance and upgrades and replacing kitchens, bathrooms and appliances."
+  >
+    <div className="text-sm text-white space-y-7 mt-7">
+      <div>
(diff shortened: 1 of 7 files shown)


app/components/ui/MaintenanceExplainerDrawer.tsx
```diff
@@ -0,0 +1,36 @@
...
+        <b>Low 1.5%</b>
+        <p>I will do the minimum to keep the home habitable, only replacing appliances or fittings when they break. Over time things will begin look tatty but the house will remain habitable.</p>
+      </div>
+      <div>
+        <b>Medium 2%</b>
+        <p>I will do enough to keep things nice, replacing things when they&apos;re old, decorating occasionally and re-fitting bathrooms and kitchens every decade or so.</p>
+      </div>
+      <div>
+        <b>High 2.5%</b>
+        <p>I will keep things nice, and also improve the property at some point during my time there.For example, extending the house or doing an energy retrofit project.</p>
```
zz-hh-aa, line 29: nit: missing space

zz-hh-aa:
Looks swish!

zz-hh-aa (approved):

PR #260, opened 2025-01-09
test: test class calculations

zz-hh-aa:
Previously our tests only covered instantiation (eg `expect(fairholdLandPurchase).toBeInstanceOf(Fairhold)`) and didn't actually test any of the computed properties. 

This PR creates tests for those and uses the test helpers created in #214 to instantiate the classes.


DafyddLlyr (approved):
Great!

PR #263, opened 2025-01-09
nit: fix placeholder copy in form

zz-hh-aa:
NB: we are using 0 for newbuild age now

Q: see comment in line 330 (should we hook up the mortgage variable explainers to the actual constants so we don't have to maintain them in two places?)

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -6,6 +6,7 @@ import { Household } from "@/app/models/Household";
 import Dashboard from "./Dashboard";
 import { formSchema, FormFrontend } from "@/app/schemas/formSchema";
 import { useSearchParams } from "next/navigation";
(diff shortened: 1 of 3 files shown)


DafyddLlyr (approved):
Nice catches here thanks!

re: Maintaining the variables - yes, ideally we'd import them in here (they're coming from `/models` right?) - this is a great shout.

zz-hh-aa:
Just before I merge-- @DafyddLlyr in new commit [8db84c1](https://github.com/theopensystemslab/fairhold-dashboard/pull/263/commits/8db84c1031b347dab444fbf111247f65e4e72331) do you think it makes sense that I moved the mortgage variables from `Mortgage.ts` to `constants.ts` since they now need exporting / are used in multiple places? I always looked for them in the wrong place before anyway (always though they'd be in `contants.ts`...)

DafyddLlyr (approved):
Perfect @zz-hh-aa ! 🙌

PR #264, opened 2025-01-09
feat: Setup final card

DafyddLlyr:
Screenshot 2025-01-09 at 18 22 58

A little rough and ready (styles aren't perfect) but the content and functionality is all here. As with other PRs, links are currently empty.

PR #265, opened 2025-01-09
feat: Initial setup of "what difference..." card (5/6)

DafyddLlyr:
As with #264, this isn't perfect in terms of styles but the layout is pretty close and we can have a pass over with font sizes/colours etc when a few more things come together 👌 

This does _not_ work on mobile - focusing on rough layout / content here at this stage over this.

image


gabrielegranello (approved):
Looks very nice!

PR #268, opened 2025-01-10
feat: update maintenance levels

### LINKED ISSUE #223: Update `MAINTENANCE_LEVELS` constant (and thus `HOUSE_BREAKDOWN_PERCENTAGES`) with updated figures

zz-hh-aa:
While starting on calculating `Lifetime` for all maintenance levels, I realised it would make more sense to update them all to the latest version now than when there are even more instances!

Closes #223

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -6,7 +6,7 @@ import { Household } from "@/app/models/Household";
 import Dashboard from "./Dashboard";
 import { formSchema, FormFrontend } from "@/app/schemas/formSchema";
 import { useSearchParams } from "next/navigation";
-import { DEFAULT_INTEREST_RATE, DEFAULT_MORTGAGE_TERM, DEFAULT_INITIAL_DEPOSIT } from "../../models/constants"
+import { DEFAULT_INTEREST_RATE, DEFAULT_MORTGAGE_TERM, DEFAULT_INITIAL_DEPOSIT, MAINTENANCE_LEVELS } from "../../models/constants"
 
 import { RadioGroup, RadioGroupItem } from "@/components/ui/radio-group";
 import { ClipLoader } from "react-spinners";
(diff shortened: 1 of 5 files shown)


app/components/ui/CalculatorInput.tsx
```diff
@@ -52,7 +52,7 @@ const CalculatorInput = () => {
       maintenancePercentage: [0.015, 0.02, 0.0375].includes(
```
gabrielegranello, line 52: Should 0.02 become 0.019 here too? Can I ask you why 0.019? It seems a weird number compared to 0.015 and 0.02 but maybe there is a good reason for it?

zz-hh-aa: Great catch!! Yes that should have been replaced, I went with the option of replacing all hard-coded maintenance level figures with referencing the constant (so I needed to fix some types too). That should be all updated with commit [7ff6db2](https://github.com/theopensystemslab/fairhold-dashboard/pull/268/commits/7ff6db24e93ac59ccd8baa118221dcdff40f2fa9)

gabrielegranello (approved):

app/components/ui/CalculatorInput.tsx
```diff
@@ -280,41 +280,41 @@ const CalculatorInput = () => {
...
+                          value={MAINTENANCE_LEVELS[2].toString()}
                           id="radio-option-high"
                           className="radio-button-style"
                         />
                         <Label
                           htmlFor="radio-option-high"
                           className="radio-label-style"
                         >
-                          High (3.75%)
+                          High ({MAINTENANCE_LEVELS[2] * 100}%)
```
DafyddLlyr, line 317: Maybe we should model these as an object, not an array.

```suggestion
                          High ({MAINTENANCE_LEVEL["high"] * 100}%)
```

app/models/ForecastParameters.ts
```diff
@@ -13,7 +13,7 @@ export interface ForecastParameters {
  * all values except years are percentages represented in decimal form
  */
 export const DEFAULT_FORECAST_PARAMETERS: ForecastParameters = {
-  maintenancePercentage: 0.02, 
+  maintenancePercentage: 0.019,
```
DafyddLlyr, line 16:
```suggestion
  maintenancePercentage: MAINTENANCE_LEVELS["medium"]
```

DafyddLlyr (commented):
One suggestion about how we store and use these constants but not a showstopper!

app/components/ui/CalculatorInput.tsx
zz-hh-aa: Good shout--just started on that before realising it will make my next PR more complicated 😅 so jut going to make a card for this and move on for now!

PR #269, opened 2025-01-10
feat: tenure cost per month

gabrielegranello:
# What does this PR do?
Adds the content to the second card. Still work in progress..


app/components/graphs/TenureComparisonBarChart.tsx
```diff
@@ -1,24 +1,28 @@
...
+    label: "Freehold",
+    color: "rgb(var(--freehold-land-color-rgb))",
+  },
+  privateRent: {
+    label: "Private Rent",
+    color: "rgb(var(--private-rent-color-rgb))",
+  },
+} satisfies ChartConfig;
+
+import React from "react";
```
DafyddLlyr, line 25: All imports should be grouped at the top of the file.

app/components/graphs/TenureComparisonBarChart.tsx
```diff
@@ -38,86 +43,68 @@ interface StackedBarChartProps {
...
+            <XAxis
+              dataKey="tenure"
+              tickLine={false}
+              tickMargin={10}
+              axisLine={false}
+              tickFormatter={(value) => value}
+            />
+            <ChartTooltip content={<ChartTooltipContent hideLabel />} />
+            <ChartLegend content={<ChartLegendContent />} />
+            <Bar dataKey="monthly" strokeWidth={2} activeIndex={2} />
```
DafyddLlyr, line 102: We want to embed the `LabelList` component here (please see #272)

DafyddLlyr (changes requested):

app/components/graphs/TenureComparisonBarChart.tsx
gabrielegranello: Thanks! Not sure if you agree, but I slightly change the format function otherwise it was showing 
4k, 2k and 630 when looking at monthly values.

gabrielegranello: Screenshot 2025-01-10 at 17 12 29

app/lib/format.ts
DafyddLlyr, line 1: Good call - this makes more sense given the context of the second graph 👍

app/components/Dashboard/Cards/HowMuchPerMonth.tsx
```diff
@@ -0,0 +1,26 @@
+import GraphCard from "../../ui/GraphCard";
+import TenureComparisonWrapper from "../../graphs/TenureComparisonWrapper";
+import { Drawer } from "../../ui/Drawer";
+import { Household } from "@/app/models/Household";
+
+interface DashboardProps {
```
DafyddLlyr, line 6: This looks like a repeated interface - each card that takes in data will be the same. Could we share this or make a common component for our cards?

app/components/graphs/TenureComparisonBarChart.tsx
```diff
@@ -0,0 +1,106 @@
...
+  affordabilityMonthly: number;
+  [key: string]: string | number;
+};
+
+// Define the props type for StackedBarChart
+interface StackedBarChartProps {
+  data: DataInput[];
+}
+
+// Implementation of the Chart.js Stacked Bar Chart
```
DafyddLlyr, line 32: nit: Redundant / noisy comments

app/components/graphs/TenureComparisonBarChart.tsx
```diff
@@ -0,0 +1,106 @@
...
+      monthly: data[0].socialRent + data[1].socialRent,
+      fill: "rgb(var(--social-rent-color-rgb))",
+    },
+  ];
+
+  return (
+    <Card>
+      <CardHeader></CardHeader>
+      <CardContent>
+        <ChartContainer config={{}}>
```
DafyddLlyr, line 78: nit: Empty config?

app/components/graphs/TenureComparisonWrapper.tsx
```diff
@@ -0,0 +1,78 @@
+"use client";
+import React from "react";
+import ErrorBoundary from "../ErrorBoundary";
+import { Household } from "@/app/models/Household";
+import TenureComparisonBarChart from "./TenureComparisonBarChart";
+
+interface TenureComparisonWrapperProps {
```
DafyddLlyr, line 7: Are these props correct? We're only passing a household into this wrapper.

DafyddLlyr (changes requested):

app/components/graphs/TenureComparisonBarChart.tsx
gabrielegranello: It seems that `config` is a compulsory variable to pass. We used to pass some properties in terms of colours for the plot. However, the same properties can be passed `chartData`. Since we actually don't need to pass any other `config` parameter, but we need to pass something otherwise it flags an error, I thought the solution is to pass an empty objet. Is  there perhaps a better way?

app/components/Dashboard/Cards/HowMuchPerMonth.tsx
gabrielegranello: Good point! I will have `DahsboardProps` exported when it is declared, so it can be imported on all the modules

app/components/graphs/TenureComparisonWrapper.tsx
gabrielegranello: Good point thanks for flagging it

app/components/graphs/TenureComparisonBarChart.tsx
DafyddLlyr: Ah didn't know this - if that's the case then this is great 👌

DafyddLlyr (approved):

PR #286, opened 2025-01-16
feat: store `MAINTENANCE_LEVELS` as object, not array

### LINKED ISSUE #274: Store `MAINTENANCE_LEVELS` constant as an object (with level keys) instead of array

zz-hh-aa:
# What does this PR do?
- Makes the `MAINTENANCE_LEVELS` constant an object, not array; and exports a new type `MaintenanceLevel` for the keys
- Correctly accesses the new data structure across the codebase
- Deletes an unused component from an earlier UI version 
    - When I was updating all instances of `MAINTENANCE_LEVELS` I found this unused component; following the principle of deleting unused code ([Slack thread](https://opensystemslab.slack.com/archives/C01RF806126/p1736511804409129)) because we have access to Git history, I've opted for deleting it (since it's not in the current Miro design anyway) and it meant one less thing to fix.  

# Why?
While starting on #251, I realised that we should update the data type of the `MAINTENANCE_LEVELS` constant first (before having to update it in more places). 

This also just makes code more legible and useful in cases where we want the level itself (eg `low` and now the value eg `0.015`).  

Closes #274


app/components/ui/CalculatorInput.tsx
```diff
@@ -51,9 +49,13 @@ const CalculatorInput = () => {
...
           : "D", // Default value for house type
-        maintenancePercentage: urlMaintenancePercentage && (MAINTENANCE_LEVELS as readonly number[]).includes(Number(urlMaintenancePercentage))
-          ? Number(urlMaintenancePercentage) as z.infer<typeof maintenancePercentageSchema>
-          : MAINTENANCE_LEVELS[1],
+        maintenanceLevel: 
+          urlMaintenanceLevel === "high"  ||
+          urlMaintenanceLevel === "medium"  ||
+          urlMaintenanceLevel === "low"  ||
+          urlMaintenanceLevel === "none"
+            ? urlMaintenanceLevel
```
DafyddLlyr, line 57:
```suggestion
        maintenanceLevel: ["high", "medium", "low", "none"].includes(urlMaintenanceLevel)
            ? urlMaintenanceLevel
```

app/models/Property.ts
```diff
@@ -40,7 +38,7 @@ export class Property {
    * Size of the house in square meters
    */
   size: number;
-  maintenancePercentage: MaintenancePercentage;
+  maintenanceLevel: 'none' | 'low' | 'medium' | 'high';
```
DafyddLlyr, line 41:
```suggestion
  maintenanceLevel: MaintenanceLevel
```

app/models/Property.ts
```diff
@@ -112,8 +110,10 @@ export class Property {
     componentKey: keyof houseBreakdownType,
     newBuildPrice: number,
     age: number,
-    maintenanceLevel: number
+    maintenanceLevel: keyof typeof MAINTENANCE_LEVELS
```
DafyddLlyr, line 113:
```suggestion
    maintenanceLevel: MaintenanceLevel
```

DafyddLlyr (approved):
Nice 👌 Much more readable and consistent

PR #287, opened 2025-01-17
chore: allow House age to be 0

gabrielegranello:
# What does this PR do?
It changes the validation schema to allow for a value equal to 0 in the House Age field. This is because now a new house has year equal to 0 and not equal to 1.

### DIFF
--- a/app/schemas/calculationSchema.ts
+++ b/app/schemas/calculationSchema.ts
@@ -1,23 +1,25 @@
 import { z } from "zod";
-import {
-  parse as parsePostcode,
-  isValid as isValidPostcode,
-} from "postcode";
+import { parse as parsePostcode, isValid as isValidPostcode } from "postcode";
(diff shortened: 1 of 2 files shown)


app/schemas/calculationSchema.ts
```diff
@@ -30,7 +28,9 @@ export const calculationSchema = z.object({
     .transform(parsePostcode)
     .refine((postcode): postcode is ValidPostcode => postcode.valid),
   houseSize: z.coerce.number().positive("houseSize must be a positive integer"),
-  houseAge: z.coerce.number().positive("houseAge must be a positive integer"),
+  houseAge: z.coerce
+    .number()
+    .nonnegative("houseAge must be a positive integer or 0"),
```
gabrielegranello, line 33: Sorry I am not sure what prettier is doing, but I literally only change this line in calculationSchema and the equivalent line in formSchema

DafyddLlyr (approved):

Issue #270, opened 2025-01-10
Make floor area optional


zz-hh-aa:
- Make optional in frontend
- Create default value in backend
- If no value provided, map bedroom # to area using default values (Alastair will produce these)

![Image](https://github.com/user-attachments/assets/30d581d5-817b-4f2b-b90f-8727e9526168)

PR #288, opened 2025-01-17
feat: Optional floor area

gabrielegranello:
# What does this PR do?
Following #270 , it allows `houseSize` in the input to be undefined, so that it becomes an optional parameter. If `undefined`, the value is then mapped to the number of bedrooms.

#How?
The value is made optional in the `formSchema` and `calculationSchema`. Note that `houseSize` is not necessary to fetch the data in the API. Later, when the `Property` class is actually constructed, there is an if condition that checks if `houseSize` exists or needs to be mapped.

#Anything else?
We used to have a file named `testClasses.ts` but it was not a test. The main function within it was called `calculateFairhold`. I think the name was confusing, so I renamed it `calculateFairhold.ts`. Last, I added a series on unit tests in `calculateFairhold.test.ts`

### DIFF
--- a/app/models/calculateFairhold.ts
+++ b/app/models/calculateFairhold.ts
@@ -1,4 +1,4 @@
-import { ValidPostcode } from './../schemas/calculationSchema';
+import { ValidPostcode } from "../schemas/calculationSchema";
 import { createForecastParameters } from "./ForecastParameters";
 import { Household } from "./Household";
 import { HouseType, MaintenancePercentage, Property } from "./Property";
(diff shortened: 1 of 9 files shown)


app/models/calculateFairhold.ts
```diff
@@ -35,13 +35,27 @@ function calculateFairhold(responseData: ResponseData) {
     throw new Error("maintenancePercentage data is missing or empty");
   }
 
+  function assignHouseSize(numberOfBedrooms: number) {
```
DafyddLlyr, line 38: I'd view this functionality as part of our data validation, which is being handled by zod.

It's acceptable for the API to receive `houseSize` as undefined, but if it does it should map to a number based on the supplied number of rooms.

This approach would maintain the separation between data validation and business logic.

DafyddLlyr (changes requested):

PR #293, opened 2025-01-21
feat: new `maintenanceCost` and `depreciatedHouseResaleValue` objects in `Lifetime`

zz-hh-aa:
# What does this PR do?
Makes previous properties into objects to tidy up the `Lifetime` class

# Why?
Having multiple similar properties seemed verbose (eg `maintenanceCostLow`, `maintenanceCostMedium`, `maintenanceCostHigh`) but also placing them in objects with `MaintenanceType` type keys will make them easier to access elsewhere. This came up when working on graph 2, which is why I'm doing it now!


DafyddLlyr (approved):
Nice - a big improvement!

PR #296, opened 2025-01-24
feat: add gas object to Household

zz-hh-aa:
When working on social value, I found we needed to access the kwh demand figure from `Household`, so this PR refactors `Household` to store those values in a new `gasDemand` object property and makes its methods more concise too.

### DIFF
--- a/app/models/Household.ts
+++ b/app/models/Household.ts
@@ -38,8 +38,12 @@ export class Household {
     fairholdLandRent: FairholdLandRent;
   };
   public lifetime: Lifetime;
-  public gasBillExistingBuildYearly: number;
-  public gasBillNewBuildOrRetrofitYearly: number;
+  public gasDemand = {
+    kwhExistingBuildYearly: 0,
+    kwhNewBuildOrRetrofitYearly: 0,
+    billExistingBuildYearly: 0,
+    billNewBuildOrRetrofitYearly: 0
+  }
 
   constructor(params: ConstructorParams) {
     this.incomePerPersonYearly = params.incomePerPersonYearly;
@@ -47,8 +51,7 @@ export class Household {
     this.property = params.property;
     this.forecastParameters = params.forecastParameters;
     this.incomeYearly = HEADS_PER_HOUSEHOLD * params.incomePerPersonYearly;
-    this.gasBillExistingBuildYearly = this.calculateGasBillExistingBuild(params);
-    this.gasBillNewBuildOrRetrofitYearly = this.calculateGasBillNewBuildOrRetrofit(params);
+    this.gasDemand = this.calculateGasDemand(params)
     this.tenure = this.calculateTenures(params);
     this.lifetime = this.calculateLifetime(params);
   }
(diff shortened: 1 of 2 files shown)


app/models/Household.ts
```diff
@@ -142,13 +145,24 @@ export class Household {
...
+    private calculateKwh(params: ConstructorParams, newOrExisting: string) {
+      const kwhM2 = newOrExisting === "existingBuild" ? KWH_M2_YR_EXISTING_BUILDS : KWH_M2_YR_NEWBUILDS_RETROFIT
+      const kwhYearly = params.property.size * kwhM2[params.property.houseType]
+      return kwhYearly
+    }
+    
+    private calculateGasBill(newOrExisting: string) {
+      const kwhYearly = newOrExisting === "existingBuild" ? this.gasDemand.kwhExistingBuildYearly : this.gasDemand.kwhNewBuildOrRetrofitYearly
+      const gasBillExistingBuildYearly = this.kwhCostPence * kwhYearly / 100
+      return gasBillExistingBuildYearly
```
DafyddLlyr, line 166: I found this quite hard to read / quite dense.

We always need to calculate both of these values together for both `calculateGasBill()` and `calculateKwh()`.

It might be simpler to do them in a single function rather than have a conditional based on `newOrExisting`;

```ts
private calculateGasBill() {
      const existing = this.kwhCostPence * this.gasDemand.kwhExistingBuildYearly / 100;
      const newBuild = this.kwhCostPence * this.gasDemand.kwhNewBuildOrRetrofitYearly / 100;

      return { exiting, newBuild }
}
```
      
Alternatively these could be very small and simple individual functions.

DafyddLlyr (changes requested):

app/models/Household.ts
zz-hh-aa: Yep thanks--had just ended up starting a refactor on a child branch just as you were reviewing! How is the version in [519c282](https://github.com/theopensystemslab/fairhold-dashboard/pull/296/commits/519c282a109a0e1a36406703137bf83318f5edca)?

app/models/Household.ts
```diff
@@ -38,17 +38,20 @@ export class Household {
...
   };
   public lifetime: Lifetime;
-  public gasBillExistingBuildYearly: number;
-  public gasBillNewBuildOrRetrofitYearly: number;
+  public gasDemand = {
+    kwhExistingBuildYearly: 0,
+    kwhNewBuildOrRetrofitYearly: 0,
+    billExistingBuildYearly: 0,
+    billNewBuildOrRetrofitYearly: 0
+  }
```
DafyddLlyr, line 46: Does this have to be an instantiated object, or can it just be a type?

DafyddLlyr (commented):
Changes are much clearer! One small question / comment.

app/models/Household.ts
zz-hh-aa: Heh this was me confusing myself 😓 changed back to type in [eca2ba6](https://github.com/theopensystemslab/fairhold-dashboard/pull/296/commits/eca2ba6f9be1aa96d2746a4ba6c72c058cce8923)

DafyddLlyr (approved):

PR #297, opened 2025-01-24
feat: social value calculations

zz-hh-aa:
- Creates a new `SocialValue` class within `Household` (and associated tests)
- Renders `SocialValue` figures in the `WhatDifference` card

### DIFF
--- a/app/components/Dashboard/Cards/WhatDifference.tsx
+++ b/app/components/Dashboard/Cards/WhatDifference.tsx
@@ -1,6 +1,7 @@
 import { Household } from "@/app/models/Household";
 import { Drawer } from "../../ui/Drawer";
 import GraphCard from "../../ui/GraphCard"
+import { SOCIAL_VALUE_YEARS } from "@/app/models/constants";
 
 type CardsProps = {
   household: Household;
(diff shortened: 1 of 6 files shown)


app/models/constants.ts
```diff
@@ -177,4 +177,10 @@ export const KWH_M2_YR_NEWBUILDS_RETROFIT = {
   T: 20,
   S: 51,
   D: 55
-}
\ No newline at end of file
+}
+
+export const KG_CO2_PER_KWH = 0.185 // from Carbon Independent https://www.carbonindependent.org/15.html#:~:text=Natural%20gas&text=Older%20gas%20meters%20measure%20gas,kg%20%2F%20kWh%20%5B9%5D%20.
+
+export const NHS_SAVINGS_PER_HEAD_PER_YEAR = 24.78 // from BRE report in 2021, NHS first-year care costs due to poor housing in England individualised
```
DafyddLlyr, line 184: If you change these to JSDoc comments, IDEs will pick them up which is a nice DX improvement ✨ 

image

app/models/SocialValue.test.ts
DafyddLlyr, line 1: Great to see these here 💪

app/models/SocialValue.ts
```diff
@@ -0,0 +1,64 @@
+import { Household } from "./Household"
+import { KG_CO2_PER_KWH, NHS_SAVINGS_PER_HEAD_PER_YEAR, FTE_SPEND } from "./constants";
+
+type ConstructorParams = {
+    household: Household};
```
DafyddLlyr, line 5:
```suggestion
type ConstructorParams = {
    household: Household
};
```

app/models/SocialValue.ts
```diff
@@ -0,0 +1,64 @@
...
+    public embodiedCarbonSavings: number;
+    public savingsEnergyPoundsYearly: number;
+    public savingsToNHSPerHeadYearly: number;
+    public localJobs: number;
+    public operationalCarbonSavingsYearly: number;
+
+    constructor(params: ConstructorParams) {
+        this.moneySaved = this.calculateMoneySavedFHLP(params);
+        this.communityWealthDecade = this.calculateCommunityWealth(params);
+        this.embodiedCarbonSavings = 31.89; // TODO: update figures, not placing in constants.ts because it's placeholder; static number comparing average brick & block emissions vs. timber on slab
```
DafyddLlyr, line 19: Super helpful comment 👍

app/models/SocialValue.ts
```diff
@@ -0,0 +1,64 @@
...
+            marketPurchaseTotal += (lifetime[i].marketLandMortgageYearly + lifetime[i].newbuildHouseMortgageYearly)
+            fairholdLandPurchaseTotal += (lifetime[i].fairholdLandMortgageYearly + lifetime[i].depreciatedHouseMortgageYearly)
+        }
+        const moneySaved = marketPurchaseTotal - fairholdLandPurchaseTotal
+        return moneySaved;
+    }
+    private calculateCommunityWealth(params: ConstructorParams) {
+        const lifetime = params.household.lifetime.lifetimeData
+        let communityWealth = 0
+        for (let i = 0; i < 10; i++) { // TODO: decide on time period
```
DafyddLlyr, line 40: Maybe we should encode this "10" as a variable?

app/models/SocialValue.ts
```diff
@@ -0,0 +1,64 @@
...
+        const operationalEmissionsNewBuildOrRetrofit = KG_CO2_PER_KWH * params.household.gasDemand.kwhNewBuildOrRetrofitYearly
+        const operationalEmissionsSavedTCo2e = (operationalEmissionsExistingBuild - operationalEmissionsNewBuildOrRetrofit) / 1000
+        return operationalEmissionsSavedTCo2e
+    }
+
+    private calculateLocalJobsSupported(params: ConstructorParams) {
+        const maintenanceLevel = params.household.property.maintenanceLevel
+        const totalSpend = params.household.property.newBuildPrice + params.household.lifetime.lifetimeData[0].maintenanceCost[maintenanceLevel]
+        let jobsSupported = totalSpend / FTE_SPEND
+        jobsSupported = parseFloat(jobsSupported.toFixed(1));
```
DafyddLlyr, line 61: Is this right or do we want whole numbers only? "4.5 jobs" is a bit odd (I guess it's full time equivalent though right?)

DafyddLlyr (approved):
A few question or nitpicks, but all good 👌

app/components/Dashboard/Cards/WhatDifference.tsx
```diff
@@ -27,43 +35,43 @@ const Highlight: React.FC<React.PropsWithChildren> = ({ children }) => (
   <span className="text-green-500 font-semibold">{children}</span>
 )
 
-const Cards = () => (
+const Cards: React.FC<CardsProps> = ({ household }) => (
   <div className="flex flex-wrap gap-6">
-    <Card title="Money saved" figure="£30,000">
-      <p>On housing costs over 10 years, compared to conventional ownership</p>
+    <Card title="Money saved" figure={`£${Math.round(household.socialValue.moneySaved).toLocaleString()}`}>
```
DafyddLlyr, line 40: If you're not familiar with it, have a look at this API - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat

app/components/Dashboard/Cards/WhatDifference.tsx
```diff
@@ -27,43 +35,43 @@ const Highlight: React.FC<React.PropsWithChildren> = ({ children }) => (
...
+      <p>If Fairhold Land Purchase, on housing costs over 10 years compared to conventional ownership</p>
     </Card>
     <Card title="Community wealth">
       <p>
         Every 10 years the house would contribute{" "}
-        <Highlight>£120,000</Highlight> to community infrastructure and services
+        <Highlight>£{Math.round(household.socialValue.communityWealthDecade).toLocaleString()}</Highlight> to community infrastructure and services
```
DafyddLlyr, line 46: nit: It might be more readable to encode these as variables first rather than doing it all in-line.

app/models/SocialValue.ts
zz-hh-aa: Yeah, FTE--will make that clearer in the copy. Following Alastair's designs here image

DafyddLlyr: Perfect - thanks for clarifying that this is the intention here.

PR #305, opened 2025-01-28
feat: side drawer 'Find out more...' content

zz-hh-aa:
# What's new in this PR
- Installs `react-markdown`
- New directory `Dashboard/Help/` that stores the drawer content in markdown files, named identically to the cards themselves
- Updates the cards to use the markdown content
    - Some of the cards use string processing and `.replace()` in order to not hard-code variables 
- I also had to move a tiny bit of math out of a constant and into the `Lifetime` class

# Why
- Maintaining separate markdown files with the side drawer content in is easier than dealing with HTML tags etc., thanks to Daf for [pattern suggestion](https://opensystemslab.slack.com/archives/C01RF806126/p1738062818961619?thread_ts=1738062572.110359&cid=C01RF806126)

# Questions
❓ Is there a tidier way of going about rendering Markdown in the drawer?
❓ Most of this PR is Markdown (which is why there are so many lines), should it be split up into a few smaller ones?

### DIFF
--- a/app/components/Dashboard/Cards/CostOverTime.tsx
+++ b/app/components/Dashboard/Cards/CostOverTime.tsx
@@ -4,8 +4,11 @@ import CostOverTimeWrapper, { TenureType } from "../../graphs/CostOverTimeWrappe
 import { Drawer } from "../../ui/Drawer";
 import TenureSelector from "../../ui/TenureSelector";
 import { DashboardProps } from "../../ui/Dashboard";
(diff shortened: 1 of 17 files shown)


zz-hh-aa:
Note to self to update the NHS save help text after the chain of #308, #307 and #300 merges

gabrielegranello (approved):
Ok I see, I think I should have reviewed this one before the other one. I can see a few things that will be changed later in terms of maths and replacement.

PR #307, opened 2025-01-29
feat: calculate and render cumulative cost by tenure

zz-hh-aa:
# What does this PR do?
- Adds an object to `Lifetime` to iterate through and store cumulative spend by tenure
- Accesses the figures from `CostOverTime` card
    - Updates the colour according to tenure selected
    - Changes the `GraphCard` component to accept a component _or_ just a string (I wasn't so sure about this move, but thought it ultimately makes sense since other subtitles are straightforward strings?)

Now the subtitle updates the figure and colour according to selected tenure
image

image

# Why?
As shown in the designs:
image

Alastair also explained the colour / figure changing functionality [in Slack](https://opensystemslab.slack.com/archives/C01RF806126/p1738103335315569?thread_ts=1738067412.188359&cid=C01RF806126)

### DIFF
--- a/app/components/ui/GraphCard.tsx
+++ b/app/components/ui/GraphCard.tsx
@@ -2,7 +2,7 @@ import React from "react";
 
 type Props = React.PropsWithChildren<{
   title: string;
-  subtitle?: string;
+  subtitle?: string | React.ReactNode;
 }>;
 
 const GraphCard: React.FC<Props> = ({ title, subtitle, children }) => {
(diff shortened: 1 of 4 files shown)


app/components/ui/GraphCard.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 
 type Props = React.PropsWithChildren<{
   title: string;
-  subtitle?: string;
+  subtitle?: string | React.ReactNode;
```
DafyddLlyr, line 5:
```suggestion
  subtitle?: React.ReactNode;
```

Checkout the type definition of `React.ReactNode` and see if this throws any errors. I'm pretty sure that this would be valid and correct.

DafyddLlyr (approved):
Looking good!

app/components/ui/GraphCard.tsx
zz-hh-aa: Yep good catch, worked perfectly :)

PR #316, opened 2025-02-07
chore: rename testClasse file

gabrielegranello:
# What does this PR do?
It renames the `testClasses.ts` file into `calculateFairhold.ts`.

# Why?
I think the name `testClasses.ts` is misleading because no test is involved. The file contains the function calculateFairhold, so maybe we should change its name?

I was thinking actually a follow up PR should be developing the unit testing for it, i.e. `calculateFairhold.test.ts` because it is currently missing. OThinking of a file named `testClasses.test.ts` will become even more confusing to me


zz-hh-aa (approved):
Looks good, thanks for getting round to this chore!

PR #317, opened 2025-02-07
feat: optional floor area

gabrielegranello:
# What does this PR do?
It allows to treat the floor area as optional.

# How?
By two steps:
1) In the form validation schema, we allow to submit an empty string. If empty, the associated value will be null
2) In the calculation schema, if the value is empty, we apply the mapping function

# Note:
I had to explicitly tell zod:
.transform((val) => (val === "" ? undefined : Number(val)))
how do deal with an empty string because it did not like the case where first a negative number was inserted, and then submitted with empty string.

### DIFF
--- a/app/api/route.test.ts
+++ b/app/api/route.test.ts
@@ -1,6 +1,6 @@
 import { POST } from "../api/route";
 import * as calculationService from "../services/calculationService";
-import calculateFairhold from "../models/testClasses";
+import calculateFairhold from "../models/calculateFairhold";
 import { NextResponse } from "next/server";
 
 // Mock dependencies
--- a/app/api/route.test.ts
+++ b/app/api/route.test.ts
@@ -5,7 +5,7 @@ import { NextResponse } from "next/server";
 
 // Mock dependencies
 jest.mock("../services/calculationService");
-jest.mock("../models/testClasses", () => jest.fn()); // Mock calculateFairhold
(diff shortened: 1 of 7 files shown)


zz-hh-aa:
Thanks Gabry! 

I checked out this branch and tested it out, got a couple questions / comments.

1.  When I didn't enter a house size in the frontend, running it threw an error: 
``` 
ERROR: API -  [
  {
    "code": "invalid_type",
    "expected": "number",
    "received": "nan",
    "path": [
      "houseSize"
    ],
    "message": "Expected number, received nan"
  }
]
``` 
Does it work fine for you? 

2. I'm also not seeing a change in the frontend marking the field as optional, could you please add this in?

3. One more question: I'm wondering why the assigning default value happens in `calculationSchema` rather than in the `Property` class, for instance (which is where I would have put it instinctively). Not asking for this change, just wondering about your process :)

gabrielegranello:
Hi Ollie thanks for your feedback, appreciated! Now hopefully it should be fixed, there was something weird. I have added the optional text in the frontend and quick styling class, is that ok?
About where to place that chunk of code, I think we think alike  :) 
That was my first attempt but then Daf made an excellent point about splitting logic from form validation. You can read his comments here #288

zz-hh-aa (approved):
Thanks Gabry, looking good / working well for me! Leaving an approval but if you want to wait for Daf's eyes before merging, no prob :)

PR #318, opened 2025-02-07
feat: restructure depreciation to be incremental & store house breakdown by component each year

zz-hh-aa:


gabrielegranello (approved):
Can't find any issues, looks good!

PR #323, opened 2025-02-11
feat: safe string replace when rendering markdown

zz-hh-aa:
# What does this PR do?
- Installs relevant dependencies (`remark` for string processing and `unis-util-visit` which is used to iterate through the 'nodes' of markdown content and replace values where needed)
- In the relevant cards, Use `remark` to replace previous usage of `.replace` method
- New file `Dashboard/Cards/types.ts` with the `TextNode` type (to resolve type issue)

# Why?
- Doing string processing directly on the markdown files was vulnerable to injection attacks

### DIFF
--- a/app/components/Dashboard/Cards/CostOverTime.tsx
+++ b/app/components/Dashboard/Cards/CostOverTime.tsx
@@ -61,7 +61,7 @@ export const CostOverTime: React.FC<DashboardProps> = ({ processedData }) => {
     };
     
     processMarkdown();
-  });
+  }, [replacements]);
 
   return (
     <GraphCard
--- a/app/components/Dashboard/Cards/CostOverTime.tsx
+++ b/app/components/Dashboard/Cards/CostOverTime.tsx
@@ -1,4 +1,4 @@
-import { useState } from "react";
+import { useState, useEffect, useMemo } from "react";
 import GraphCard from "../../ui/GraphCard";
 import CostOverTimeWrapper, { TenureType } from "../../graphs/CostOverTimeWrapper";
 import { Drawer } from "../../ui/Drawer";
@@ -9,6 +9,9 @@ import ReactMarkdown from 'react-markdown';
 import explanationContent from '../Help/CostOverTime.md';
 import { DEFAULT_FORECAST_PARAMETERS } from "@/app/models/ForecastParameters";
 import { SOCIAL_RENT_ADJUSTMENT_FORECAST } from "@/app/models/constants";
+import { remark } from "remark";
+import { visit } from 'unist-util-visit';
+import type { TextNode } from "./types";
 
 const TENURES = ['marketPurchase', 'marketRent', 'fairholdLandPurchase', 'fairholdLandRent', 'socialRent'] as const
 const TENURE_LABELS = {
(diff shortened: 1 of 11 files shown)


app/components/Dashboard/Cards/CostOverTime.tsx
```diff
@@ -19,25 +22,36 @@ const TENURE_LABELS = {
...
+            });
+            node.value = value;
+          });
+        })
+        .process(explanationContent);
+      setProcessedContent(result.toString());
+    };
+    
+    processMarkdown();
+  }, [replacements]);
```
gabrielegranello, line 54: Should useEffect be triggered when `explanationContent` changes rather than `replacements`? 
My understanding is that `replacements` is inside `useMemo`, so that should not change, should it?

app/components/Dashboard/Cards/HowMuchPerMonth.tsx
```diff
@@ -1,27 +1,48 @@
...
+            });
+            node.value = value;
+          });
+        })
+        .process(explanationContent);
+      setProcessedContent(result.toString());
+    };
+    
+    processMarkdown();
+  }, [replacements]);
```
gabrielegranello, line 44: Same as above

app/components/Dashboard/Cards/ResaleValue.tsx
```diff
@@ -16,13 +19,32 @@ interface DashboardProps {
...
+            });
+            node.value = value;
+          });
+        })
+        .process(explanationContent);
+      setProcessedContent(result.toString());
+    };
+    
+    processMarkdown();
+  }, [replacements]);
```
gabrielegranello, line 47: same as above

gabrielegranello (changes requested):
Couple of small tweaks , otherwise looking good to me. Great work!

app/components/Dashboard/Cards/CostOverTime.tsx
```diff
@@ -19,25 +22,36 @@ const TENURE_LABELS = {
...
-
   const [selectedTenure, setSelectedTenure] = useState<TenureType>('marketPurchase');
+  const [processedContent, setProcessedContent] = useState('');
+  
+  // We don't hard code the variables in markdown because then we'd have to maintain them in multiple places
+  const replacements = useMemo(() => ({
+    constructionPriceGrowthPerYear: (DEFAULT_FORECAST_PARAMETERS.constructionPriceGrowthPerYear * 100).toString(),
+    rentGrowthPerYear: (DEFAULT_FORECAST_PARAMETERS.rentGrowthPerYear * 100).toString(),
+    propertyPriceGrowthPerYear: (DEFAULT_FORECAST_PARAMETERS.propertyPriceGrowthPerYear * 100).toString(),
+    incomeGrowthPerYear: (DEFAULT_FORECAST_PARAMETERS.propertyPriceGrowthPerYear * 100).toString(),
```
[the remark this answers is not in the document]
zz-hh-aa: Great catch, thanks!! 👁️

app/components/Dashboard/Cards/CostOverTime.tsx
zz-hh-aa: Excellent catch, I've removed that now across all three instances: [ebb328c](https://github.com/theopensystemslab/fairhold-dashboard/pull/323/commits/ebb328cdab38221de69592fa7d906c83c8958348)

gabrielegranello: Sorry do we not need [explanationContent] as a dependency? My understanding is that useEffect should be triggered when `explanationContent` changes. If we pass an empty function, does it mean it will be triggered at every re-render of the page?

zz-hh-aa: Ok I did some more research and this is the context as I understand it:
1. `useEffect` will always run at least once when the component mounts
2. An empty array should mean it runs just once; it's what I had on my last fix but then I got a lint warning suggesting that I remove it, so I did
3. `explanationContent` is a static Markdown file and so it's a constant; there's not really any point in listing it as a dependency
4. Since ESLint wants the array to have content, it's okay to put `replacements` in there (it won't cause extra re-renders since it's memoized)

I have reverted back to listing `replacements` as a dependency to resolve the lint issue and prevent unnecessary re-rendering (which not including an array would lead to) in [d736621](https://github.com/theopensystemslab/fairhold-dashboard/pull/323/commits/d7366216940e6b09e54bb65e6c51c91f9ae2e29b)

gabrielegranello (approved):

PR #328, opened 2025-02-13
feat: differentiate purchase price and rents by bedrooms

zz-hh-aa:
# What does this PR do?
- Weight purchase price by beds (price paid dataset only differentiates by property type; bed weighting developed following [this source](https://www.dwh.co.uk/advice-and-inspiration/average-house-sizes-uk) for average beds by property type)
- Migration for adding a `bedrooms` column to the rent table
- Migration to insert rental data by beds 
- Update the rent repo / service to access data by beds too
- 
# Why? 
We wanted to explore why monthly freehold costs were so much higher than private rent costs, and see if adjusting purchase prices and rents by number of bedrooms would lessen the difference. This was largely the case, though sometimes the difference was slight. 
image

# Questions
❓ Is a big long `INSERT` the best way to handle adding a bunch of data to the `rent` table (in addition to the `TRUNCATE`)
❓ The rental data only goes up to 4 bedrooms, while our other weights go up to 6; is it worth adding weighting here?


gabrielegranello (approved):
Nicely done.

PR #329, opened 2025-02-14
feat: create filtered prices paid summary table (filter out outliers)

zz-hh-aa:
What does this PR do?
- Filters prices paid data by area (eg SE) by first and last quartiles

Why?
Following [Gabry's suggestion](https://opensystemslab.slack.com/archives/C01RF806126/p1739524249696609?thread_ts=1739465942.321799&cid=C01RF806126), trying to create slightly more realistic data.

While sometimes the difference between freehold and private rent is even greater than in the other two methods, ultimately I think this method leads to the closest figures. 

This branch uses `oz/weight-purchase-price` as a base, so the property weightings and rental data by number of bedrooms are also inputs. 

image

### DIFF
--- a/app/data/pricesPaidRepo.ts
+++ b/app/data/pricesPaidRepo.ts
@@ -16,7 +16,7 @@ const getPricesPaidByPostcodeAndHouseType = async (
   houseType: string
 ): Promise<PricesPaidParams> => {
   try {
-    const summary = await prisma.pricesPaidSummary.findFirst({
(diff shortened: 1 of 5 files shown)


zz-hh-aa:
As discussed with Alastair, this seems like a reasonable approach to take. No need to graph distribution for now since eliminating outliers is an established stats method.

gabrielegranello (approved):
Looks great to me. Great work!

PR #331, opened 2025-02-14
chore: tidy graphs

### LINKED ISSUE #330: Tidy graphs
- [ ] Make sure axis titles are uniform
- [ ] Tidy tooltip text
- [ ] Swap order of land-house tooltip to match graph
Image
- [ ] Make units uniform

zz-hh-aa:
# What does this PR do?
- Renames graphs and their wrappers so that they match their content card names in `Dashboard/`
- Removes tooltip from graph 2 because I thought it was redundant (❓ unsure about this one, what do people think?? The old tooltip label just showed what was already in the bar)
image
- Changes tooltip in graph 1 to match styling in 3 and 4 (and to show 'Total' instead of redundant information)
Before: 
Screenshot 2025-02-14 143700
After: 
Screenshot 2025-02-14 143706
- Makes axes naming and styling consistent in graphs 3 and 4

Closes #330

### DIFF
--- a/app/components/graphs/HowMuchFHCostBarChart.tsx
+++ b/app/components/graphs/HowMuchFHCostBarChart.tsx
@@ -11,19 +11,15 @@ import { NameType } from "recharts/types/component/DefaultTooltipContent";
 
 const chartConfig = {
   freeholdLand: {
-    label: "Land ",
     color: "rgb(var(--freehold-equity-color-rgb))",
   },
   freeholdHouse: {
-    label: "House ",
     color: "rgb(var(--freehold-interest-color-rgb))",
   },
   fairholdLand: {
-    label: "Land ",
     color: "rgb(var(--fairhold-equity-color-rgb))",
   },
   fairholdHouse: {
-    label: "House ",
     color: "rgb(var(--fairhold-interest-color-rgb))",
   },
 } satisfies ChartConfig;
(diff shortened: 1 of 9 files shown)


app/components/graphs/HowMuchFHCostBarChart.tsx
```diff
@@ -1,29 +1,29 @@
...
-  ChartTooltip,
-  ChartTooltipContent,
 } from "@/components/ui/chart";
+import { ValueType } from "tailwindcss/types/config";
+import { NameType } from "recharts/types/component/DefaultTooltipContent";
 
 const chartConfig = {
   freeholdLand: {
-    label: "Land",
+    label: "Land ",
```
DafyddLlyr, line 14: Is there a way we could introduce this whitespace via margin/padding instead of a `" "` character. It looks like a mistake a well-intentioned dev might tidy up in future.

DafyddLlyr (approved):
Great!

app/components/graphs/HowMuchFHCostBarChart.tsx
zz-hh-aa: Good catch, thanks! I didn't realise these labels are redundant anyway 'cause I removed them in favour of showing the total in the tooltip (the breakdown figures are already shown in the graph)

PR #342, opened 2025-02-27
feat: add affordability threshold to monthly cost graph

zz-hh-aa:
Adds the 35% of household income line
image

following designs
image


gabrielegranello (approved):
Nice one!

PR #348, opened 2025-02-28
Gg/dropdown

gabrielegranello:
# What does this PR do?
It installs & adds the dropdown with the correct style. It substitutes #310 

https://github.com/user-attachments/assets/caca3015-17d6-426f-a717-3b21ae504ba4

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -30,6 +30,7 @@ import { Label } from "@/components/ui/label";
 import { APIError } from "@/app/lib/exceptions";
 import { BackgroundAssumptions } from "./BackgroundAssumptions";
 import { MaintenanceExplainerDrawer } from "./MaintenanceExplainerDrawer";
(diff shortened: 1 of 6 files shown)


app/components/ui/CalculatorInput.tsx
```diff
@@ -233,11 +248,11 @@ const CalculatorInput = () => {
                       How old is the building?
```
zz-hh-aa, line 248: Little copy nit: question should now be "When was the house built?"

app/globals.css
```diff
@@ -81,6 +81,23 @@ body {
   font-size: 12px;
 }
 
+.dropdown-style {
+  background: rgb(var(--background-input-field-rgb)) !important;
```
zz-hh-aa, line 85: Did all of these `!important`s end up being necessary to preserve the correct styling?

zz-hh-aa (changes requested):

zz-hh-aa (approved):
Nice, thank you :)

PR #351, opened 2025-03-05
feat: use median GDHI data instead of average

zz-hh-aa:
# What does this PR do
- Migration to drop all previous GDHI data, replace `itl3` column with `itl1` and insert new median GDHI data
- Update GDHI repo and service accordingly 
- Update `Household` class (eg we no longer need per-head income, `HEADS_PER_HOUSEHOLD` constant and a household income property; we just retrieve the one household income figure and save to `Household.incomeYearly` 

# Why
Our previous GDHI data was using average income per-head and multiplying this by the average household size. This led to higher and less accurate household income figures. 

# To-dos after merging
- Update the help text on how this was calculated (see PR #323)
- Update sources (`gdhi` script in [fairhold-data](https://github.com/theopensystemslab/fairhold-data) becomes redundant; source should also be updated to [IFS](https://ifs.org.uk/data-items/median-household-incomes-region))
- (Maybe) adjust discount formula accordingly as affordability figures  / inputs will be different now


gabrielegranello (approved):
Looks good to me, well done

PR #353, opened 2025-03-05
feat: set max y axis for monthly costs

zz-hh-aa:
Turns out it wasn't a bug 😅 

# What does this PR do?
- Sets the y axis of the monthly costs graph to be the maximum of monthly market mortgage cost, monthly Fairhold cost, or 35% affordability threshold (rounded to nearest £500)

# Why?
Realised that the affordability line wasn't showing in areas where land values are negative. 
Before:
image

After:
image

### DIFF
--- a/app/components/graphs/HowMuchPerMonthBarChart.tsx
+++ b/app/components/graphs/HowMuchPerMonthBarChart.tsx
@@ -1,7 +1,7 @@
 "use client";
 
 import React from "react";
-import { Bar, BarChart, CartesianGrid, XAxis, LabelList, ReferenceLine } from "recharts";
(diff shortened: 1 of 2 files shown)


gabrielegranello (approved):
I like the plots without the Y axis, they seem more tidy to me. However, it is hard to then read the affordability line value.
Proposal: should we explicitly the value in the grey line? Something like 35% threshold equal to £1234 in another PR?

zz-hh-aa:
> I like the plots without the Y axis, they seem more tidy to me. However, it is hard to then read the affordability line value. Proposal: should we explicitly the value in the grey line? Something like 35% threshold equal to £1234 in another PR?

No Y axis is closer to the original design as well! Will run the £1234 option past Alastair

PR #355, opened 2025-03-06
feat: bottom out money saved at 0

zz-hh-aa:
In cases where a Fairhold home would be more expensive than market (eg newbuild in an area with negative land values / where the market price of a home is less than construction cost)

Previous:
image


gabrielegranello (commented):
Do you think we should max it out at 0 or keep the negative number but in red? Should we also add a commentary message?: For example, it looks like Fairhold is not convenient in your area. This is most likely because etc etc?

zz-hh-aa:
> Do you think we should max it out at 0 or keep the negative number but in red? Should we also add a commentary message?: For example, it looks like Fairhold is not convenient in your area. This is most likely because etc etc?

Alastair suggested bottoming out at 0. Haven't discussed other text options but can always add later?

gabrielegranello (approved):

PR #356, opened 2025-03-06
test: raise test coverage threshold and cover missing lines

### LINKED ISSUE #276: Look into test coverage issue
We lowered the test coverage threshold in #247 as a temporary fix for this

Daf in Slack convo:
>Not sure what’s happening with a number of PRs failing for 100 coverage when there’s no related changes - it’s a bit flakey and will pass/fail on different runs with no changes.
In order to keep up momentum this week, we can maybe drop to 95% or something and then turn our attention back to this later?

zz-hh-aa:
In #247 we lowered test coverage thresholds temporarily. This PR raises the threshold and improves test coverage. 

Closes #276


gabrielegranello (approved):
I think it's fine. I also get that some lines are not covered but maybe we can save this for later if needed? I'd agree it is not on top of our priority list

PR #361, opened 2025-03-10
Gg/UI form

gabrielegranello:
# What does this PR do?
It sets the styling of UI form for different screen sizes.

Large screen
Screenshot 2025-03-10 at 10 04 18

Medium screen
Screenshot 2025-03-10 at 10 04 37

Phone
Screenshot 2025-03-10 at 10 04 45

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -131,7 +131,7 @@ const CalculatorInput = () => {
               name="houseType" // Name in the Calculation schema for the new radio field
               render={({ field }) => (
                 <FormItem>
-                  <FormLabel className="h3-style">House type</FormLabel>
+                  <FormLabel className="h3-style ">House type</FormLabel>
                   <FormControl>
                     <RadioGroup
                       value={field.value}
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -116,11 +116,11 @@ const CalculatorInput = () => {
 
   if (view === "form") {
     return (
-      <div className="flex flex-col justify-center max-w-xl mx-auto px-10">
-        <div className="h1-style text-3xl sm:text-2xl md:text-xl">
+      <div className="flex flex-col justify-center max-w-2xl mx-auto px-10">
+        <div className="h1-style text-lg md:text-xl lg:text-2xl">
           Calculate how Fairhold could work for you
         </div>
-        <div className="subheadstyle text-xl sm:text-lg md:text-base">
+        <div className="subheadstyle text-sm sm:text-sm md:text-base lg:text-base">
           Compare the estimated cost of a Fairhold home in your area with other
           ways of owning or renting.
         </div>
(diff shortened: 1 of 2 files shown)


app/components/ui/CalculatorInput.tsx
```diff
@@ -116,11 +116,11 @@ const CalculatorInput = () => {
...
   if (view === "form") {
     return (
-      <div className="flex flex-col justify-center max-w-xl mx-auto px-10">
-        <div className="h1-style">
+      <div className="flex flex-col justify-center max-w-2xl mx-auto px-10">
+        <div className="h1-style text-lg md:text-xl lg:text-2xl">
           Calculate how Fairhold could work for you
         </div>
-        <div className="subheadstyle">
+        <div className="subheadstyle text-sm sm:text-sm md:text-base lg:text-base">
```
zz-hh-aa, line 123: I think some of this might be redundant, you might want to test it out yourself but I believe `<div className="subheadstyle text-sm md:text-base">` will achieve the same effect?

app/components/ui/CalculatorInput.tsx
```diff
@@ -131,12 +131,14 @@ const CalculatorInput = () => {
               name="houseType" // Name in the Calculation schema for the new radio field
               render={({ field }) => (
                 <FormItem>
-                  <FormLabel className="h3-style">House type</FormLabel>
+                  <FormLabel className="h3-style text-sm md:text-sm lg:text-base ">
```
zz-hh-aa, line 134: Same here with double `text-sm`

app/components/ui/CalculatorInput.tsx
```diff
@@ -197,18 +199,20 @@ const CalculatorInput = () => {
...
             />
-            <div className="grid grid-cols-2 gap-4">
+            <div className="grid grid-cols-1 md:grid-cols-2 gap-4">
               <FormField
                 control={form.control}
                 name="housePostcode"
                 render={({ field }) => (
                   <FormItem>
-                    <FormLabel className="h3-style">Postcode</FormLabel>
+                    <FormLabel className="h3-style text-sm md:text-sm lg:text-base ">
```
zz-hh-aa, line 208: And here

app/components/ui/CalculatorInput.tsx
```diff
@@ -221,16 +225,18 @@ const CalculatorInput = () => {
                 name="houseSize"
                 render={({ field }) => (
                   <FormItem>
-                    <FormLabel className="h3-style">
+                    <FormLabel className="h3-style text-sm md:text-sm lg:text-base">
```
zz-hh-aa, line 228: ➕

app/components/ui/CalculatorInput.tsx
```diff
@@ -221,16 +225,18 @@ const CalculatorInput = () => {
...
                       How big is the house?{" "}
-                      <span className="optional-style">(Optional)</span>
+                      <strong className="optional-style text-xs md:text-xs lg:text-sm">
```
zz-hh-aa, line 230: ➕  and a couple more instances below I think!

zz-hh-aa (changes requested):
UI looking great, thank you! Commented some code nits.

app/components/ui/CalculatorInput.tsx
gabrielegranello: Thanks @zz-hh-aa good point! I actually started with all options and then manually adjusted the size to see which one was best. Ultimately only two styles are necessary. I will amend as per your comments

zz-hh-aa (approved):
Nice, thanks for those updates!

PR #372, opened 2025-03-11
feat: add axes and split line 35% threshold label in monthly cost graph

zz-hh-aa:
Old version:
image

New (with axes and splitting the 35% threshold label over two lines):
image

Recharts can't render JSX so we can't use `\n` or `<br>` to create a line break. I'm not sure if my solution is overly complex--couldn't figure anything else out but I have to say I don't like it very much!

### DIFF
--- a/app/components/graphs/HowMuchPerMonthBarChart.tsx
+++ b/app/components/graphs/HowMuchPerMonthBarChart.tsx
@@ -105,14 +105,32 @@ const HowMuchPerMonthBarChart: React.FC<StackedBarChartProps> = ({
               y={data[0].affordabilityMonthly} 
               stroke="rgb(var(--text-inaccessible-rgb))" 
               strokeDasharray="6 6" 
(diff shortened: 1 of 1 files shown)


app/components/graphs/HowMuchPerMonthBarChart.tsx
```diff
@@ -107,14 +105,32 @@ const HowMuchPerMonthBarChart: React.FC<StackedBarChartProps> = ({
...
               strokeDasharray="6 6" 
-              label={{ 
-                value: '35% median household income',
-                position: 'insideTopRight',
-                fill: 'rgb(var(--text-inaccessible-rgb))',
-                fontSize: 12, 
-                offset: 10
-              }} 
-            />        
+              label={(props) => { // formatting here is to ensure line break
```
gabrielegranello, line 108: Small nit, remove comments?

PR #373, opened 2025-03-11
wip: total label above bars

zz-hh-aa:
image


gabrielegranello (approved):
Great work, looks good to me. Just a question: do we want the total to be positioned at that constant hight, or do we want it just on top of the bar? (hence different heights depending on the bar height). It could be a separate PR in case

gabrielegranello:
Oh sorry I didn't realise it was a draft, please ignore me

zz-hh-aa:
> Great work, looks good to me. Just a question: do we want the total to be positioned at that constant hight, or do we want it just on top of the bar? (hence different heights depending on the bar height). It could be a separate PR in case

Alastair would like them positioned on top of the bar, but I was struggling to achieve this! Struggling with SVG layout, hence WIP. We can chat about it at dev call 🙂

PR #376, opened 2025-03-13
fix: social value constants

zz-hh-aa:
When working on the documentation for the social value section of the calculator (see [wiki](https://github.com/theopensystemslab/fairhold-dashboard/wiki)), l wasn't able to replicate some of the constants and so am updating (with docs this time..!). 

I'd appreciate another pair of eyes on how both of these figures were calculated.

### DIFF
--- a/app/models/constants.ts
+++ b/app/models/constants.ts
@@ -193,11 +193,17 @@ export const FTE_SPEND = 60000
 
 export const SOCIAL_VALUE_YEARS = 10;
 
-/** from Leeds Beckett Wikihouse LCA comparison report */
-export const EMBODIED_CARBON_BRICK_BLOCK_KG_M2 = 404.086
+/** 
+ * from Leeds Beckett Wikihouse LCA comparison report: 37.58 tco2e for a 92m2 brick and block home  
+ * https://cdn.prod.website-files.com/612569c3243035968aaf61f1/63bd27536d1e152e28e96403_77084_Wikihouse%20Thermal%20Perfromance%20and%20LCA%20report_July%202021.pdf
+*/
+export const EMBODIED_CARBON_BRICK_BLOCK_KG_M2 = 408.478
 
-/** from Stride Treglown Gwynfaen embodied emissions estimates */
-export const EMBODIED_CARBON_TIMBER_SLAB_KG_M2 = 59.933
+/** 
+ * from Stride Treglown Gwynfaen embodied emissions estimates: 40,017kg up-front CO2e for life cycle stages A1-A5 and biogenic carbon sequestration of 36,177kg A1-A3 for a 95m2 home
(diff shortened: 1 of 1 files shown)


app/models/constants.ts
```diff
@@ -193,11 +193,17 @@ export const FTE_SPEND = 60000
 
 export const SOCIAL_VALUE_YEARS = 10;
 
-/** from Leeds Beckett Wikihouse LCA comparison report */
-export const EMBODIED_CARBON_BRICK_BLOCK_KG_M2 = 404.086
+/** 
+ * from Leeds Beckett Wikihouse LCA comparison report: 37.58 tco2e for a 92m2 brick and block home  
+ * https://cdn.prod.website-files.com/612569c3243035968aaf61f1/63bd27536d1e152e28e96403_77084_Wikihouse%20Thermal%20Perfromance%20and%20LCA%20report_July%202021.pdf
```
gabrielegranello, line 198: Sorry @zz-hh-aa can you point me please to what page, section, paragraph this number comes from? I was searching for it at page 46, Section 3, Table 3.11 The LCA of the Wikihouse Skylark and alternative cavity wall building (kgCO2-eq / 100 years) and I see different numbers, but not sure if I am looking in the right place.

app/models/constants.ts
```diff
@@ -193,11 +193,17 @@ export const FTE_SPEND = 60000
...
+*/
+export const EMBODIED_CARBON_BRICK_BLOCK_KG_M2 = 408.478
 
-/** from Stride Treglown Gwynfaen embodied emissions estimates */
-export const EMBODIED_CARBON_TIMBER_SLAB_KG_M2 = 59.933
+/** 
+ * from Stride Treglown Gwynfaen embodied emissions estimates: 40,017kg up-front CO2e for life cycle stages A1-A5 and biogenic carbon sequestration of 36,177kg A1-A3 for a 95m2 home
+ * https://media.stridetreglown.com/wp-content/uploads/2021/03/23114044/152858_LCA-Case-Study_Gwynfaen-final.pdf
+*/
+export const EMBODIED_CARBON_TIMBER_SLAB_KG_M2 = 40.421
```
gabrielegranello, line 206: Same, I was looking at page 24, section 8.5. Superstructure comparison - Embodied carbon over the life cycle.
I also noticed that report one considers emissions over a life of 100 years, while the second for 60 years. Do we need to decide on one of the 2, and convert one into the other?

app/models/constants.ts
zz-hh-aa: I think this should be 30.66 actually, just using A1:A5 since B will be covered by the operational emissions section (we're not calculating life cycles here). Do you think that makes sense?

app/models/constants.ts
zz-hh-aa: For the sake of consistency we should use A1:A5 and so the time question becomes irrelevant, right? As B (operational) is what's time-dependent.

PR #393, opened 2025-03-18
feat: allow user to enter half (outward) postcode only

### LINKED ISSUE #385: Can we allow half postcodes only? eg for users who want to test what a home in another location would cost but don't know an exact postcode

zz-hh-aa:
# What
- New type called `PostcodeScales` that allows `null`s for certain values, should work if user enters outward postcode instead of full 
    - Uses type in schema
- Updates tests

# Why
Alastair raised this in roadmap: what happens if someone wants to test a dream home location but doesn't know a full postcode? 

# Questions
I'm not sure if this was the most elegant way of doing this--suggestions very much appreciated. 

Closes #385

### DIFF
--- a/app/api/route.test.ts
+++ b/app/api/route.test.ts
@@ -58,15 +58,12 @@ describe("POST API Route", () => {
       ...validApiInput,
       // Parsed postcode object
       housePostcode: {
+        outcode: "SE17",
+        incode: "1PE",
         area: "SE",
         district: "SE17",
-        incode: "1PE",
-        outcode: "SE17",
-        postcode: "SE17 1PE",
         sector: "SE17 1",
-        subDistrict: null,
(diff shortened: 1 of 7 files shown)


zz-hh-aa:
Made some changes following our call earlier today. Seems like previous casts `as string` in `calculationSchema` lines 50-56 were because `parsePostcode` returns either a `ValidPostcode` _or_ `InvalidPostcode` (in which all values are `null`). 

The `postcode` library does not have a type for partial postcodes, which is why the `PostcodeScales` type with partially `null`able values was necessary. 

With [089c911](https://github.com/theopensystemslab/fairhold-dashboard/pull/393/commits/089c91165def25ff6d3af51e5df1b0cc8436d5a6), the db query (with possible `null` sector) is now updated.

PR #397, opened 2025-03-28
feat: fix and style error messages

### LINKED ISSUE #384: Make error text red and write more meaningful error messages
Replace with "Enter a valid postcode" or similar--service design principle: don't say please!

Image
Image

zz-hh-aa:
Ensures that meaningful error messages show (following content design principles (eg "Do this"), styled red to be more visible.

image

Closes #384 (where the 'before' screenshots can be found)

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -101,7 +101,7 @@ const CalculatorInput = () => {
       case "INSUFFICIENT_PRICES_PAID_DATA":
         form.setError("housePostcode", {
           message:
-            "Insufficient data for this postcode. Please try again with a different postcode",
+            "Insufficient data for this postcode. Please try again with a different postcode.",
         });
         break;
       case "UNHANDLED_EXCEPTION":
@@ -195,7 +195,7 @@ const CalculatorInput = () => {
                       </div>
                     </RadioGroup>
                   </FormControl>
-                  <FormMessage />
+                  <FormMessage className="--error-text-rgb" />
                 </FormItem>
               )}
             />
@@ -215,7 +215,7 @@ const CalculatorInput = () => {
                         className="inputfield-style text-xs"
                       />
                     </FormControl>
-                    <FormMessage />
+                    <FormMessage  className="error-message-style" />
                   </FormItem>
                 )}
               />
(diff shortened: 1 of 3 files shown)


gabrielegranello (changes requested):
Screenshot 2025-04-04 at 14 29 37
I get the following error when selecting any value. I think the issue is that in the frontend, we assign houseAge to be a `number`. In the schema, we have zod raising an error if that value is NOT a `string`

zz-hh-aa:
Thanks @gabrielegranello great catch! I got so tied up in testing errors I forgot to test if correct values would still work... fix should be up at [3605279](https://github.com/theopensystemslab/fairhold-dashboard/pull/397/commits/36052796dc36cbe773a2f1f9d92d0d7c19926693)

gabrielegranello:
Screenshot 2025-04-04 at 14 58 57
Sorry I think we need to cover the case for New. I think in zod we are forcing the number to be greater than 0, which is incompatible with New houses

zz-hh-aa:
omg I'm getting careless 😓 thanks for the eagle eyes @gabrielegranello! Newbuilds _and_ existing builds should be working now, see [534d282](https://github.com/theopensystemslab/fairhold-dashboard/pull/397/commits/534d282cbfeae940f23bbb4908cb3d921c0ccd12).

gabrielegranello:
Sorry my bad, I think I did not explain the error properly

https://github.com/user-attachments/assets/27ca5ba8-d840-49db-ac69-cc2c751c7977

gabrielegranello:
Got it. I think the issue is at line 
234 in CalculatorInput
It should be maybe: onValueChange={(value) => field.onChange(value)}?
Otherwise I don't think that we are changing the value itself but only the label?

zz-hh-aa:
Really strange, thanks for looking into it--I can't seem to replicate it though, have I missed something? 

https://github.com/user-attachments/assets/72dfdfc8-552a-4127-a40c-a5bd0b845f64

gabrielegranello (approved):
Sorry maybe it was cache issue on my end. This morning everything was running fine, sorry about that

zz-hh-aa:
> Sorry maybe it was cache issue on my end. This morning everything was running fine, sorry about that

No problem, appreciate you testing so thoroughly!

PR #415, opened 2025-04-03
feat: responsive graphs dashboard

### LINKED ISSUE #360: UI: Responsive styles - graph dashboard layout

zz-hh-aa:
# What does this PR do?
- Responsive styling in the graphs: just looking at layout for the moment, not addressing all of Alastair's new changes in [Figma](https://www.figma.com/design/SThTPsRJbUzhgPJIgjIXdo/Fairhold-Calculator?node-id=0-1&t=Ow50PrRJhYO13L7A-1)
- New CostOverTimeStackedBarChartMobile component (this is the only one that needs to rotate on mobile), conditional logic / new `UseScreenSize` hook will select which one to render
-  `GraphCard` component now has responsive titles / subtitles

Closes #360

### DIFF
--- a/app/components/Dashboard/Cards/CostOverTime.tsx
+++ b/app/components/Dashboard/Cards/CostOverTime.tsx
@@ -75,7 +75,7 @@ export const CostOverTime: React.FC<DashboardProps> = ({ processedData }) => {
         </span>
       }
       >
-      <div className="flex flex-col h-full w-3/4 justify-between">
+      <div className="flex flex-col h-full w-full md:w-3/4 justify-between">
         <div className="flex gap-2 mb-4">
         {TENURES.map(tenure => (
           <TenureSelector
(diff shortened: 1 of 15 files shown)


gabrielegranello (commented):
There is something bizarre going on with  the Social rent button.. Sometimes it  is not clicking when I am with the cursor exactly on top of the text. I had a look at the code but for what I can see all the buttons are treated the same so I cannot explain myself why. Maybe is is because I am using dev mode in the browser?

https://github.com/user-attachments/assets/5f39484d-ede3-470b-a52d-e725b92c947c

[if you turn the volume up you can actually hear me clicking)

gabrielegranello:
Unfortunately I cannot check the deployed version on Vercel, but I wonder if it makes sense to test there? The code seems correct to me

gabrielegranello (approved):

PR #421, opened 2025-04-07
feat: restructure and restyle social value

zz-hh-aa:
# What does this PR do? 
Restyles the social value card (`WhatDifference`) and makes some minor updates to the `SocialValue` class since we decided to show benefits over `lifetime` length and not the somewhat arbitrary 10 year perios. 

# Questions
In line 56, the effects of `flex-col` and `flex-row` seem to be reversed. Am I being stupid? I removed styling from parent and grandparent etc. elements to try to isolate the issue and wasn't able to, so just kept the classes as is to achieve desired look. 
___

I'd say the bulk of the restyling is done, this is what desktop looks like:
image

And responsive sizes etc are working on mobile (scrolling / snap is a separate issue so I would like to keep that in a separate PR):
image

Design for reference:
image

### DIFF
--- a/app/components/Dashboard/Cards/WhatDifference.tsx
+++ b/app/components/Dashboard/Cards/WhatDifference.tsx
@@ -1,7 +1,7 @@
 import { Household } from "@/app/models/Household";
 import { Drawer } from "../../ui/Drawer";
 import GraphCard from "../../ui/GraphCard"
-import { SOCIAL_VALUE_YEARS, DEFAULT_INTEREST_RATE, DEFAULT_MORTGAGE_TERM, DEFAULT_INITIAL_DEPOSIT } from "@/app/models/constants";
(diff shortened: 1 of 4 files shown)


gabrielegranello:
I think this is line 56 right?
`<div className="flex flex-col sm:flex-row gap-6 w-3/4 justify-center py-4">`
My understanding of tailwind is that it has the mobile first approach. Therefore, on mobile it will be flex-col by default. For a small screen size and bigger, it will switch to flex-row https://tailwindcss.com/docs/responsive-design
does it help?
Screenshot 2025-04-14 at 14 41 25

gabrielegranello:
There is something strange in the mobile version. The last card about taking the survey somehow disappears. Is it just me?

https://github.com/user-attachments/assets/b23a5e2b-5432-4211-8635-d3019fd2b85c

gabrielegranello (approved):

PR #427, opened 2025-04-11
feat: restyle up-front cost graph

### LINKED ISSUE #413: Add % of market price to up-front cost graph subtitle
Image

zz-hh-aa:
- Adds totals on the top left of each bar (uses the parent element to position, thanks Daf for the suggestion!)
- Adds Fairhold as % of market in subtitle (with coloured highlight)
- Updates tooltip to show value breakdown based on what part of bar is being hovered over
    - Re-styles tooltip
    - Removes values from chart label
- Fixes mobile width
- Adds colours to bar labels (on x axis)

Mobile:
image

Desktop:
image

Tooltip:
image

Closes #413


gabrielegranello (approved):
Very nice work :)

PR #428, opened 2025-04-11
feat: restyle monthly costs graph

### LINKED ISSUE #405: UI: Simplify affordability threshold and add explainer / label
Image

zz-hh-aa:
- Moves labels on top of bars
- Colour codes bar labels and x-axis labels too
- Makes affordability threshold line and text black and solid
- Updates copy
- Adds 35% explainer to help text button

Desktop
image

Mobile
image

Closes #405

### DIFF
--- a/app/components/graphs/HowMuchPerMonthBarChart.tsx
+++ b/app/components/graphs/HowMuchPerMonthBarChart.tsx
@@ -149,10 +149,50 @@ const HowMuchPerMonthBarChart: React.FC<StackedBarChartProps> = ({
               <Bar dataKey="monthly" strokeWidth={2} activeIndex={2}>
               <LabelList
                 dataKey="monthly"
-                position="center"
-                formatter={formatValue}
-                fill="white"
-                fontSize={12}
+                position="top"
+                content={(props) => {
+                  if (!props.x || !props.y || !props.value || props.index === undefined ) return null;
+
+                  const xPos = typeof props.x === 'number' ? props.x - 2 : 0;
+                  const yPos = typeof props.y === 'number' ? props.y - 30 : 0;
+                  const value = typeof props.value === 'number' ? props.value : parseFloat(props.value as string);
+                  const formattedValue = formatValue(value);
+
+                  const labelColor = (() => {
+                    switch (chartData[props.index].tenure) {
+                      case "Freehold":
+                        return "rgb(var(--freehold-equity-color-rgb))";
+                      case "Private Rent":
+                        return "rgb(var(--private-rent-land-color-rgb))";
+                      case "Fairhold - Land Purchase":
+                        return "rgb(var(--fairhold-equity-color-rgb))";
+                      case "Fairhold - Land Rent":
+                        return "rgb(var(--fairhold-equity-color-rgb))";
+                      case "Social Rent":
+                        return "rgb(var(--social-rent-land-color-rgb))";
+                      default:
+                        return "#666";
+                    }
+                  })();
+
+                  return (
+                    <foreignObject x={xPos} y={yPos} width={100} height={30}>
+                      <div
+                        style={{
+                          position: "absolute",
+                          left: 0,
+                          top: 0,
+                          color: labelColor,
+                          fontSize: "18px",
+                          fontWeight: 600,
+                          whiteSpace: "nowrap",
+                        }}
+                      >
+                        {formattedValue}
+                      </div>
+                    </foreignObject>
+                  );
+                }}
               />
             </Bar>
             <ReferenceLine 
--- a/app/components/graphs/HowMuchPerMonthBarChart.tsx
+++ b/app/components/graphs/HowMuchPerMonthBarChart.tsx
@@ -85,6 +85,22 @@ const HowMuchPerMonthBarChart: React.FC<StackedBarChartProps> = ({
               interval={0} 
               height={60} 
               tick={({ x, y, payload }) => {
+                const labelColor = (() => {
+                  switch (payload.value) {
+                    case "Freehold":
+                      return "rgb(var(--freehold-equity-color-rgb))";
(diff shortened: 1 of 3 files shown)


zz-hh-aa:
> Sometimes the first label disappears for large values. Do you get the same or is it just me_ 

Great catch, I adjusted the y axis of the graph and larger values should work now in [23ae379](https://github.com/theopensystemslab/fairhold-dashboard/pull/428/commits/23ae37941731ac5a13576b1fb7d38fac22d51198) but please do test and see if you come across cases that I didn't!

image

app/components/graphs/HowMuchPerMonthBarChart.tsx
```diff
@@ -132,37 +149,67 @@ const HowMuchPerMonthBarChart: React.FC<StackedBarChartProps> = ({
...
+                      case "Private Rent":
+                        return "rgb(var(--private-rent-land-color-rgb))";
+                      case "Fairhold - Land Purchase":
+                        return "rgb(var(--fairhold-equity-color-rgb))";
+                      case "Fairhold - Land Rent":
+                        return "rgb(var(--fairhold-equity-color-rgb))";
+                      case "Social Rent":
+                        return "rgb(var(--social-rent-land-color-rgb))";
+                      default:
+                        return "#666";
```
DafyddLlyr, line 174: It's probably worth encoding these once in a single place if possible.

app/components/graphs/HowMuchPerMonthWrapper.tsx
```diff
@@ -23,7 +23,7 @@ const HowMuchPerMonthWrapper: React.FC<HowMuchPerMonthWrapperProps> = ({
     household.forecastParameters
       .affordabilityThresholdIncomePercentage || 0
   const highestValue = Math.max(fairholdLandPurchaseMonthly, marketPurchaseMonthly, affordabilityMonthly)
-  const maxY = Math.ceil(highestValue / 500) * 500
+  const maxY = Math.ceil(1.1 * highestValue / 500) * 500
```
DafyddLlyr, line 26: Magic number! I guess the 1.1 allows for some headroom?

app/components/graphs/HowMuchPerMonthBarChart.tsx
```diff
@@ -132,37 +149,67 @@ const HowMuchPerMonthBarChart: React.FC<StackedBarChartProps> = ({
...
-                fill="white"
-                fontSize={12}
+                position="top"
+                content={(props) => {
+                  if (!props.x || !props.y || !props.value || props.index === undefined ) return null;
+
+                  const xPos = typeof props.x === 'number' ? props.x - 2 : 0;
+                  const yPos = typeof props.y === 'number' ? props.y - 30 : 0;
+                  const value = typeof props.value === 'number' ? props.value : parseFloat(props.value as string);
+                  const formattedValue = formatValue(value);
```
DafyddLlyr, line 159: Is there a reason that these values could be undefined or not a number?

It looks like this is something we could handle upstream better (fix/tighten types, return 0 as a default) as opposed to at this granular level.

app/components/graphs/HowMuchPerMonthWrapper.tsx
zz-hh-aa: Yes that's right. If it's just a scaling thing and isn't going to be used anywhere else, is it fine to leave as a magic number? Or would adding a variable eg `const scalingFactor = 1.1` be preferable?

app/components/graphs/HowMuchPerMonthBarChart.tsx
zz-hh-aa: Good shout :) done in [cf06030](https://github.com/theopensystemslab/fairhold-dashboard/pull/428/commits/cf06030d58f7f979ba36dbf93689c743425bb8dd)

app/components/graphs/HowMuchPerMonthBarChart.tsx
zz-hh-aa: From what I understand, Recharts will pass props that are typed as potentially undefined (even if our upstream types state that numbers will be provided) in order to handle edge cases. 

I tried being stricter about types with `(props: { x: number; y: number; value: number; index: number; })` but got the following error because of type mismatches:
```
Type '(props: { x: number; y: number; value: number; index: number; }) => React.JSX.Element' is not assignable to type 'ContentType | undefined'.
  Type '(props: { x: number; y: number; value: number; index: number; }) => React.JSX.Element' is not assignable to type '(props: Props) => ReactNode'.
    Types of parameters 'props' and 'props' are incompatible.
      Type 'Props' is not assignable to type '{ x: number; y: number; value: number; index: number; }'.
        Types of property 'x' are incompatible.
          Type 'string | number | undefined' is not assignable to type 'number'.
            Type 'undefined' is not assignable to type 'number'.
```

Have I misunderstood what you're saying? Is there another solution?

zz-hh-aa: Tried out [c53431a](https://github.com/theopensystemslab/fairhold-dashboard/pull/428/commits/c53431afd4a6f7d7d4b962aeb81ca6c4f1df0672) after the call, do you think it's satisfactory?

app/components/graphs/HowMuchPerMonthWrapper.tsx
DafyddLlyr: I think it's always best to be explicit here, I'd go for a variable 👍

app/components/graphs/HowMuchPerMonthBarChart.tsx
DafyddLlyr: This is great 👍 We're still casting which is not ideal, but this may be necessary and it results in a much cleaner and clearer output.

PR #442, opened 2025-04-16
feat: restyle choice card

zz-hh-aa:
- Replaces `GraphCard` in `WhatWouldYouChoose` with JSX, since this card required different formatting to the other ones
- Rounds the 'Take the survey' button (also used on form view as 'Calculate' button)

Before:
image

After:
image


gabrielegranello (approved):
Looks great to me

PR #444, opened 2025-04-16
feat: header

### LINKED ISSUE #411: UI: Header
Image

zz-hh-aa:
What do people make of scroll behaviour now? Please check on desktop and mobile, of course, using Preview link below!

Closes #411


DafyddLlyr (approved):
Looking good - we might want to give it a set height, and reduce the card height the same amount so they still add up to `100vh`.

E.g., header = 10vh, card = 90vh

PR #454, opened 2025-04-22
feat: recategorise social value impacts

zz-hh-aa:
I think the recategorisation makes more sense conceptually, but that it looks weird. What do others think?
image


DafyddLlyr (approved):
Much clearer I think!

PR #457, opened 2025-04-23
style: restyle resale value card

zz-hh-aa:
- Updates subtitle
- Adds `LabelList` for line labels at the right side of graph
- Restyles tooltip as per design
- Uses tooltip pattern from `HowMuchFHCostBarChart` to only show when selected maintenance level line is hovered over

Current:
image

Before:
image

### DIFF
--- a/app/components/graphs/ResaleValueLineChart.tsx
+++ b/app/components/graphs/ResaleValueLineChart.tsx
@@ -73,16 +73,15 @@ const ResaleValueLineChart: React.FC<ResaleValueLineChartProps> = ({
           if (typeof x !== "number" || typeof y !== "number") return null;
 
           const label = chartConfig[dataKey].label;
+          const longestLabel = Math.max(chartConfig['none'].label.length, chartConfig['low'].label.length, chartConfig['medium'].label.length, chartConfig['high'].label.length)
           const paddingX = 8;
           const paddingY = 4;
           const fontSize = 12;
 
-          // Roughly estimate text width (monospace approximation for now)
-          const textWidth = label.length * 7;
-          const rectWidth = textWidth + paddingX * 2;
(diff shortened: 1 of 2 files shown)


app/components/graphs/ResaleValueLineChart.tsx
```diff
@@ -57,39 +52,75 @@ const ResaleValueLineChart: React.FC<ResaleValueLineChartProps> = ({
...
       strokeWidth={2}
-      strokeDasharray={dataKey === selectedMaintenance ? "0" : "5 5"}
       dot={false}
-    />
+      activeDot={false}
+      onMouseOver={() => setHoveredLine(dataKey)}
+      onMouseOut={() => setHoveredLine(null)}
+      >
+      <LabelList
+        content={({ x, y, index }) => {
```
DafyddLlyr, line 71: This is quite a lot of complexity for an inline prop, could we extract this out?

```suggestion
        content={LabelListContent}
```

DafyddLlyr (approved):
One non-blocking suggestion 👍

app/components/graphs/ResaleValueLineChart.tsx
zz-hh-aa: That's a great shout! I had a quick go and got some type issues that I couldn't figure out quickly--I also realised that this is a pattern that's reused across `ResaleValueLineChart`, `HowMuchFHCostBarChart`, `HowMuchPerMonthBarChart` and `HowMuchFHCostNotViableBarChart`, so I'm going to merge this as is and make a [new card](https://github.com/orgs/theopensystemslab/projects/5?pane=issue&itemId=109795058&issue=theopensystemslab%7Cfairhold-dashboard%7C483) to fix all instances of it in one go

PR #458, opened 2025-04-23
feat: set up survey results page

zz-hh-aa:
- Installs Airtable
- Sets up route to GET from API & send to page
- Create `SurveyGraphCard` component and relevant types
- Outlines eventual dashboard results page

Scaffolding at `/survey`:
image

Data can be viewed at `/api/survey`


app/api/survey/route.ts
```diff
@@ -0,0 +1,36 @@
+import { NextResponse } from 'next/server';
+import Airtable from 'airtable';
+
+export async function GET() {
+    const apiKey = process.env.AIRTABLE_PAT;
+    const baseId = process.env.AIRTABLE_BASE_ID;
```
gabrielegranello, line 6: do we need two different cases here is the environment  is local or production?

app/api/survey/route.ts
```diff
@@ -0,0 +1,36 @@
...
+        console.error('Missing required Airtable environment variables');
+        return NextResponse.json(
+          { error: 'Server configuration error' },
+          { status: 500 }
+        );
+      }
+    try {
+        const base = new Airtable({ apiKey }).base(baseId);
+
+        const records = await base(process.env.AIRTABLE_TABLE_ID!)
```
gabrielegranello, line 19: Does this need to be tableId? Or should we delete line 7 is tableId is not used

app/survey/types.tsx
```diff
@@ -0,0 +1,29 @@
+export type SurveyComponentProps = {
+    results: SurveyResults[];
+}
+
+export type SurveyResults = {
+    id: string;
+    uk: string;
+    nonUk: string;
+    postcode: string;
```
gabrielegranello, line 9: Should this be a postcode object or a string?

app/survey/types.tsx
```diff
@@ -0,0 +1,29 @@
...
+    ageGroup: string;
+    houseType: string;
+    currentTenure: string;
+    ownershipModel?: string | null;
```
gabrielegranello, line 13: Is there a reason why these two fields can be null?

zz-hh-aa: Not all of the questions are shown to all of the users (eg if I say I am in private rent, then I won't be asked what kind of ownership model my home uses). Maybe it's worth adding a comment to explain?

app/survey/types.tsx
zz-hh-aa: We're only asking for outward postcode in the survey, so string.

app/api/survey/route.ts
zz-hh-aa: Oo good catch 👁️ replaced with `tableId` in [d08833f](https://github.com/theopensystemslab/fairhold-dashboard/pull/458/commits/d08833fab61e9337ecac4cf2fceca5c0be5bd63b)

app/api/survey/route.ts
zz-hh-aa: I believe usingNode loads env vars according to the environment it's being run in, so the switch between local and production is automatic

gabrielegranello (approved):

PR #475, opened 2025-05-07
feat: format survey data for visualisation in backend

zz-hh-aa:
# What does this PR do?
- Create types for all graphs
- Create utility functions to take the results from Airtable and format them into Recharts-friendly shapes

# Why?
We want the backend to return data that is already formatted, and only worry about visualising it in the frontend (eg in the graph components).

# Questions
Do we want to sweat some of the answer options in the survey for more consistent types? (eg we could have question-specific types listing predefined nodes for the Sankeys.) For example, one of the questions on who a user wants to live with says "Alone" while the other one says "I live alone", and it might look neater if this was consistent. And if we do want to change the answers, how's best to handle that content-data update? I guess we'll want to create a fresh survey and fresh Airtable base.

### DIFF
--- a/app/api/survey/route.ts
+++ b/app/api/survey/route.ts
@@ -1,5 +1,7 @@
 import { NextResponse } from 'next/server';
 import Airtable from 'airtable';
+import { formatSurveyResults } from '../../survey/utils';
+import { SurveyResults } from '@/app/survey/types';
 
 export async function GET() {
     const apiKey = process.env.AIRTABLE_PAT;
@@ -19,13 +21,13 @@ export async function GET() {
         const records = await base(tableId)
         .select({})
         .all();
-        
-        const formattedRecords = records.map(record => ({
-        id: record.id,
-        ...record.fields
+
+        const results = records.map(record => ({
+          ...record.fields as SurveyResults,
         }));
-        
-        return NextResponse.json({ results: formattedRecords });
+
+        const formattedResults = formatSurveyResults(results)
+        return NextResponse.json({ formattedResults });
     } catch (error) {
         console.error('Error fetching Airtable data:', error);
         return NextResponse.json(
--- a/app/survey/types.tsx
+++ b/app/survey/types.tsx
@@ -31,16 +31,21 @@ export type SurveyResults = {
 export type FormattedSurveyResults = {
     affordFairhold: AffordFairholdResults; // doughnut
     age: AgeResults; // doughnut
-    anyMeansTenureChoice: TenureChoiceResults; // sankey
+    anyMeansTenureChoice: SankeyResults; // sankey
     country: CountryResults; // tbc
-    currentMeansTenureChoice: TenureChoiceResults; // sankey
-    houseType: HouseTypeResults; // sankey
+    currentMeansTenureChoice: SankeyResults; // sankey
+    houseType: SankeyResults; // sankey
     housingOutcomes: HousingOutcomesResults; // bar 
-    liveWith: LiveWithResults; // sankey
+    liveWith: SankeyResults; // sankey
     // postcode: ; 
     supportDevelopment: SupportDevelopmentResults; // doughnut
     supportDevelopmentFactors: SupportDevelopmentFactorsResults; // bar
     supportNewFairhold: SupportNewFairholdResults; // doughnut
+};
+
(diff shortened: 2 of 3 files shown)


app/survey/utils.tsx
```diff
@@ -0,0 +1,369 @@
...
+    HousingOutcomesResults,
+    SankeyResults,
+    SurveyResults, 
+    SupportDevelopmentResults,
+    SupportDevelopmentFactorsResults,
+    SupportNewFairholdResults,
+    FormattedSurveyResults, 
+} from "./types";
+
+export const formatSurveyResults = (results: SurveyResults[]): FormattedSurveyResults => {
```
DafyddLlyr, line 13: nit: We might be better thinking of these as raw results and results. We could end up passing around the prefix `Formatted` a whole lot when contextually it's not that relevant to the front end (which never has a distinction of raw vs formatted).

```suggestion
export const formatSurveyResults = (results: RawSurveyResults[]): SurveyResults => {
```

```suggestion
export const formatSurveyResults = (results: AirtableSurveyResults[]): SurveyResults => {
```

app/survey/utils.tsx
```diff
@@ -0,0 +1,369 @@
...
+        resultsCount: results.length,
+        supportDevelopment: formatSupportDevelopmentResults(results),
+        supportDevelopmentFactors: formatSupportDevelopmentFactorsResults(results), 
+        supportNewFairhold: formatSupportNewFairholdResults(results),
+    };
+    return formattedResults;
+}
+
+const formatAffordFairhold = (results: SurveyResults[]): AffordFairholdResults => {
+    return createPieOrBarChartData(results, "affordFairhold", [
```
DafyddLlyr, line 32: Currently, we're grabbing all record from Airtable and then aggregating each time within `createPieOrBarChartData()`.

Here's an alternative approach to consider - 
- Aggregate once for the full data set (convert `SurveyResults[]` into a single `{ name: string; value: number; }` object
- Then, just pick / filter by keys, per chart

This will help in a number of ways - 
 - Much quicker / more efficient
 - Simpler
 - Better type management without casting

app/survey/types.tsx
```diff
@@ -26,4 +26,112 @@ export type SurveyResults = {
...
+    { name: "It depends"; value: number;},
+    { name: "Quite opposed to most development"; value: number;},
+    { name: "Strongly opposed to any development"; value: number;},
+    { name: "Don't know"; value: number;}
+]
+
+export type SupportDevelopmentFactorsResults = [
+    { name: "Homes that are affordable to keyworkers"; value: number;},
+    { name: "Affordable tenure homes for low-income families"; value: number;},
+    { name: "Priority given to local residents and their families (to allow downsizing, for example)"; value: number;},
```
DafyddLlyr, line 110: General question - are these mapping of the choices written in Tally?

If so, do we need them here? If we just count all responses (regardless of actual text of the answer) that should allow us to just maintain these labels once in Tally.

DafyddLlyr (commented):
Juicy PR! Lots to talk about, left some initial comments and we can dive back in at the dev call today.

There's enough logic / complexity here that it would be worth having test coverage also 👍

app/api/survey/route.ts
```diff
@@ -19,13 +21,13 @@ export async function GET() {
         const records = await base(tableId)
         .select({})
         .all();
-        
-        const formattedRecords = records.map(record => ({
-        id: record.id,
-        ...record.fields
+
+        const results = records.map(record => ({
+          ...record.fields as SurveyResults,
```
DafyddLlyr, line 26: We can avoid casting by passing a generic type into Airtable - 

```suggestion
        const records = await base<SurveyResults>(tableId).select().all();

        const results = records.map(({ fields }) => fields);
```

Under the hood I guess it's still "casting" somewhere, but this is a bit neater and more accurate!

Note: Looks like we need to swap out `null` for `undefined` in the `SurveyResults` type 👍

app/survey/utils.tsx
zz-hh-aa: Good call, pattern followed in #485 which should replace this PR!

app/survey/utils.tsx
zz-hh-aa: Similarly done in #485

PR #476, opened 2025-05-07
feat: survey results 'Who has responded' section

zz-hh-aa:
Test at `http://localhost:3000/survey`

# What does this PR do?
Graphs for the "Who has responded?" section
- Hides postcode graph (which isn't in our wireframe / we don't know how we want to visualise this yet)
- Creates the age pie chart visualisation
- Creates the country pie chart visualisation
    - Fixes the types & helper function 

image

### DIFF
--- a/app/survey/types.tsx
+++ b/app/survey/types.tsx
@@ -67,8 +67,9 @@ export type AffordFairholdResults = [
 ]
 
 export type CountryResults = {
-    [key: string]: number;
-}
+    name: string;
+    value: number;
+}[]
 
 export type HousingOutcomesResults = [
     { name: "Security from being evicted"; value: number;},
--- a/app/survey/utils.tsx
+++ b/app/survey/utils.tsx
@@ -74,15 +74,23 @@ const formatAnyMeansTenureChoice = (results: SurveyResults[]): SankeyResults => 
 };
 
 const formatCountryResults = (results: SurveyResults[]): CountryResults => {
-    const countryData: CountryResults = {};
+    const countryCount: {[key: string]: number} = {};
+    
     results.forEach((result) => { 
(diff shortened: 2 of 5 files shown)


app/survey/utils.tsx
```diff
@@ -74,15 +74,23 @@ const formatAnyMeansTenureChoice = (results: SurveyResults[]): SankeyResults =>
 };
 
 const formatCountryResults = (results: SurveyResults[]): CountryResults => {
-    const countryData: CountryResults = {};
+    const countryCount: {[key: string]: number} = {};
```
DafyddLlyr, line 77:
```suggestion
    const countryMap = new Map<string, number>();
```

app/survey/utils.tsx
```diff
@@ -74,15 +74,23 @@ const formatAnyMeansTenureChoice = (results: SurveyResults[]): SankeyResults =>
 };
 
 const formatCountryResults = (results: SurveyResults[]): CountryResults => {
```
DafyddLlyr, line 76: nit: We're making a `countryData` object here and then reformatting to a `CountryResults` - could we save a step and generate `CountryResults` directly?

DafyddLlyr (approved):
A few suggestions / alternatives but this is great and works as expected 👍

app/survey/types.tsx
```diff
@@ -67,8 +67,9 @@ export type AffordFairholdResults = [
 ]
 
 export type CountryResults = {
-    [key: string]: number;
-}
+    name: string;
+    value: number;
+}[]
```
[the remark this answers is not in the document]
DafyddLlyr: It might actually well by the case that this format is best for recharts - in which case please ignore! (Realising this looking at #475).

PR #477, opened 2025-05-07
fix: match tenure colours to selected TenureSelector button

### LINKED ISSUE #467: Bug: TenureSelector label colours
Image

zz-hh-aa:
Fixes a styling bug. 

The issue was with the fact I was trying to dynamically create Tailwind classes using template literals, which wouldn't work at buildtime. 

Before:
image

This PR:
image

Designs:
image

Closes #467


DafyddLlyr (approved):
These static/dynamic Tailwind classes are always tricky!

PR #479, opened 2025-05-07
fix: use default low maintenance level in calculateFairhold

### LINKED ISSUE #468: Bug: Make sure house price shows default low, and not the user selected maintenance level

zz-hh-aa:
Fixes bug where the user selected maintenance level was being used to calculate `depreciatedBuildPrice` (instead of our default `low`). User selected maintenance level should only be used in the resale value highlight graph. 

Closes #468

### DIFF
--- a/app/models/Property.ts
+++ b/app/models/Property.ts
@@ -106,8 +106,7 @@ export class Property {
       const result = this.calculateComponentValue(
         key, 
         this.newBuildPrice, 
-        this.age, 
-        "low" // `calculateDepreciatedBuildPrice` is only used in `Property` and not in `Lifetime`, so we can use the default value for maintenanceLevel
+        this.age
       );
 
       depreciatedBuildPrice += result.depreciatedComponentValue;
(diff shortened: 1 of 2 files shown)


app/models/Property.ts
```diff
@@ -94,22 +97,23 @@ export class Property {
...
-    depreciatedBuildPrice = parseFloat(depreciatedBuildPrice.toFixed(PRECISION))
-  return depreciatedBuildPrice;
+    // Calculate for each component using the public method
+    for (const key of Object.keys(HOUSE_BREAKDOWN_PERCENTAGES) as (keyof houseBreakdownType)[]) {
+      const result = this.calculateComponentValue(
+        key, 
+        this.newBuildPrice, 
+        this.age, 
+        "low" // `calculateDepreciatedBuildPrice` is only used in `Property` and not in `Lifetime`, so we can use the default value for maintenanceLevel
+      );
```
DafyddLlyr, line 111: This might be better / clearer / more readable as single options object as an argument, with "low" as a default value.

```ts
// Definition
public calculateComponentValue(
  {
    componentKey,
    newBuildPrice,
    age,
    maintenanceLevel = "low",
  }: {
    componentKey: keyof houseBreakdownType;
    newBuildPrice: number;
    age: number;
    maintenanceLevel?: MaintenanceLevel;
  }
): ComponentCalculation {
...
}

// Execution
const result = this.calculateComponentValue({
  componentKey: key,
  newBuildPrice: this.newBuildPrice,
  age: this.age,
});
```

DafyddLlyr (approved):

app/models/Property.ts
zz-hh-aa: Tidy--though in doing this I realised it didn't need to be an arg anyway, since we're not reusing this function anywhere else (we can access `low` from directly within it). Done in [246332b](https://github.com/theopensystemslab/fairhold-dashboard/pull/479/commits/246332bc792400654b53dfcb7081d11a9b4e9e81)

PR #482, opened 2025-05-08
fix: always give bedrooms input a controlled value

### LINKED ISSUE #461: A component (number of bedrooms) is changing an uncontrolled input to be controlled.
https://github.com/user-attachments/assets/49620577-9691-40db-9f89-40495bbb3238

zz-hh-aa:
This is a fix for the bug that Gabry found--but it looks like it's only happening on local / development environments. On the deployed Vercel version it doesn't seem to be a problem, I checked other PR previews too. 

Unsure if this fix is still needed? It has addressed the bug on my local branches however. 

What I've done is give the bedrooms input field a default blank value so that it's always a controlled input. I made the `InputFieldProps` type an extension of `React.InputHTMLAttributes<HTMLInputElement>` so that it would take `value` as a property. 

Closes #461

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -258,6 +258,7 @@ const CalculatorInput = () => {
                       <InputField
                       placeholder="e.g. 2 for two bedrooms"
                       {...field}
+                      value={field.value?.toString() || ''}
                       error={!!fieldState.error}
                       className="inputfield-style text-xs"
                       />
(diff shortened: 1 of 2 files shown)


DafyddLlyr (approved):
This is a common React error when dealing with forms, I hit this often in PlanX land!

There's a section here on the React docs about it - https://react.dev/reference/react-dom/components/input#im-getting-an-error-a-component-is-changing-an-uncontrolled-input-to-be-controlled

It only shows on local dev mode which explains why it's not on previews 👍 

The actual cause here is that we're not providing a default value for this field - it then starts as `undefined`, gets converted to a value when an input is provided, then the error throws.

PR #484, opened 2025-05-09
feat: add ResponsiveContainer to SurveyGraphCard

zz-hh-aa:
Adds React's ResponsiveContainer to the SurveyGraphCard component so that all graphs live within a responsive size.


zz-hh-aa:
Turns out this needs to be in the graph components, not in the parent `SurveyGraphCard`.

PR #485, opened 2025-05-09
feat: aggregate survey results in backend

zz-hh-aa:
Creates an `aggregateResults` function to iterate through `RawResults` once (instead of for every single graph), following [conversation](https://github.com/theopensystemslab/fairhold-dashboard/pull/475#discussion_r2081210468) in #475. 

Since `numberResponses` isn't a key of `rawResults`, `aggregateResults` will return an object with `numberResponses` and `results` keys (aggregated results stored in `results`). 

Following [this comment](https://github.com/theopensystemslab/fairhold-dashboard/pull/475#discussion_r2081171753), original results as returned from Airtable are now `RawResults` and results that have been formatted are now just `Results` (no need for `formatted` prefix).

### DIFF
--- a/app/survey/page.tsx
+++ b/app/survey/page.tsx
@@ -12,7 +12,7 @@ import { CurrentMeansTenureChoice } from './components/graphs/CurrentMeansTenure
 import { HouseType } from './components/graphs/HouseType';
 import { HousingOutcomes } from './components/graphs/HousingOutcomes';
 import { LiveWith } from './components/graphs/LiveWith';
-import { Postcode } from './components/graphs/Postcode';
+// import { Postcode } from './components/graphs/Postcode';
 import { SupportDevelopment } from './components/graphs/SupportDevelopment';
 import { SupportDevelopmentFactors } from './components/graphs/SupportDevelopmentFactors';
 import { SupportNewFairhold } from './components/graphs/SupportNewFairhold';
@@ -73,7 +73,7 @@ export default function SurveyPage() {
                 <div className="flex flex-col md:flex-row">
                   <Country {...results} />
                   <Age {...results} />
-                  <Postcode {...results} />
+                  {/* <Postcode {...results} /> */}
                 </div>
               </div>
 
--- a/app/survey/types.tsx
+++ b/app/survey/types.tsx
@@ -33,3 +33,25 @@ export type SurveyData = {
     numberResponses: number;
     results: Results;
 }
+
+export type TickProps = {
+    textAnchor: string;
+    verticalAnchor: string;
+    orientation: string;
+    width: number;
+    height: number;
+    fontSize: number;
+    x: number;
+    y: number;
+    className: string;
+    stroke: string;
+    fill: string;
+    index: number;
+    payload: {
+        coordinate: number;
+        value: string;
+        index: number;
+        offset: number;
+    },
+    visibleTicksCount: number;
+}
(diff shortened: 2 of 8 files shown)


app/survey/page.tsx
```diff
@@ -54,50 +54,52 @@ export default function SurveyPage() {
...
               <div className="flex flex-col">
                 <h3 className="text-xl font-medium">Attitudes towards development</h3>
                 <div className="flex flex-col md:flex-row">
-                  <SupportDevelopment results={surveyData} />
-                  <SupportNewFairhold results={surveyData} />
+                  <SupportDevelopment {...results} />
+                  <SupportNewFairhold {...results} />
                 </div>
-                <SupportDevelopmentFactors results={surveyData} />
+                <SupportDevelopmentFactors {...results} />
```
DafyddLlyr, line 102: This looks like a good use case for context! ([docs](https://react.dev/learn/passing-data-deeply-with-context))

Instead of passing results into each component, we can wrap them all in a context and then access the values.

Something to try out on a future PR if you're new to this pattern 👍

app/survey/utils.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const aggregateResults = (rawResults: RawResults[]) => {
+    const results: Results = {} as Results;
+    const numberResponses = rawResults.length;
+
+    for (const rawResult of rawResults) {
+        for (const key in rawResult) {
+            if (key === "id") continue; // Skip the id field (we don't need it)
+            const validKey = key as keyof RawResults; // Explicitly assert key type
+            const name = rawResult[validKey];
```
DafyddLlyr, line 11: nit: Is `name` right here? Is this an `answer`?

app/survey/types.tsx
```diff
@@ -26,4 +22,14 @@ export type SurveyResults = {
     supportDevelopment: string;
     supportDevelopmentFactors: string[];
     supportNewFairhold: string;
-};
\ No newline at end of file
+};
+
+export type Results = Record<keyof RawResults, {
+    name: string | string[] | null | undefined;
```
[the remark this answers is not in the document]
DafyddLlyr: Actually - can this ever be `null` or `undefined`? These represent optional questions I guess right, where we may have a number of non-responses.

DafyddLlyr (commented):
`aggregateResults` is much cleaner now! A few things to look at or try - 
 - could we use helper functions to make this more readable and break it up a little?
 
 ```ts
if (Array.isArray(name)) {
    name.map((item) => addToResults(validKey, item))
} else {
    addToResults(validKey, name);
}
```

- could a `Map` data structure help us? (no need to call `find()`, methods like `get()` and `has()` are helpful)

File: app/survey/types.tsx (hunk 3 above)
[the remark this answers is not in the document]
Comment (zz-hh-aa) [reply app/survey/types.tsx:28]: Yes, that's right! Not all of the questions will be shown to users.

File: app/survey/page.tsx (hunk 1 above)
Comment (zz-hh-aa) [reply app/survey/page.tsx:102]: I'd forgotten about context!! Made a card for future PR here: #493

File: app/survey/types.tsx [hunk 4]
```diff
@@ -1,17 +1,13 @@
...
     uk: string;
     nonUk: string;
     postcode: string;
     ageGroup: string;
     houseType: string;
     currentTenure: string;
-    ownershipModel?: string | null;
-    rentalModel?: string | null;
+    ownershipModel?: string | undefined;
+    rentalModel?: string | undefined;
```
DafyddLlyr, line 10: nit: The `?` marks this as an optional property already.

PR #494, opened 2025-05-13
feat: format sankey data too

zz-hh-aa:
# What does this PR do?
Creates two sub-types for results: `BarOrPieResults` and `SankeyResults` since the data structures needed were quite different. Now the `SurveyResults` type holds `numberResponses`, `barOrPie` and `sankey`. I split them into two keys because it was becoming quite convoluted to distinguish between which properties should have the bar/pie formatting and which the sankey formatting. It was also getting needlessly complex to distinguish between which properties to overwrite or drop in the case of sankeys, where two properties from `RawResults` get aggregated into one. 

# Why? 
The original `aggregateResults()` function in #485 only formats for a bar or pie chart type, and we needed to format for sankeys too.

### DIFF
--- a/app/survey/page.tsx
+++ b/app/survey/page.tsx
@@ -3,7 +3,7 @@
 import { useState, useEffect } from 'react';
 import { Inter } from "next/font/google";
 import ErrorBoundary from '../components/ErrorBoundary';
-import { Results, SurveyData } from './types';
+import { SurveyResults } from './types';
 import { Age } from './components/graphs/Age';
 import { AffordFairhold } from './components/graphs/AffordFairhold';
 import { AnyMeansTenureChoice } from './components/graphs/AnyMeansTenureChoice';
@@ -25,7 +25,7 @@ const inter = Inter({
 });
 
 export default function SurveyPage() {
-  const [surveyData, setSurveyData] = useState<SurveyData | null>(null);
+  const [surveyData, setSurveyData] = useState<SurveyResults | null>(null);
   const [loading, setLoading] = useState(true);
   const [error, setError] = useState<string | null>(null);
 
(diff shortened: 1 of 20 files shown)


app/survey/page.tsx
```diff
@@ -25,7 +25,7 @@ const inter = Inter({
 });
 
 export default function SurveyPage() {
-  const [surveyData, setSurveyData] = useState<SurveyData | null>(null);
+  const [surveyData, setSurveyData] = useState<SurveyResults | null>(null);
```
gabrielegranello, line 28: Small nit, should these be call `surveyRetsults` and `setSurveyResults` for consistency?

app/survey/utils.tsx
```diff
@@ -1,55 +1,71 @@
-import { Results, RawResults } from "./types"
+import { RawResults, BarOrPieResults, SankeyResults } from "./types"
+
+const SANKEY_MAPPINGS = [
+  { fromKey: 'houseType', toKey: 'idealHouseType', newKey: 'houseType', isArray: false },
```
gabrielegranello, line 4: Similarly, should this key be `idealHouseType`

app/survey/utils.tsx
```diff
@@ -1,55 +1,71 @@
...
+  { fromKey: 'liveWith', toKey: 'idealLiveWith', newKey: 'liveWith', isArray: false },
```
gabrielegranello, line 5: `idealLivesWith`?

app/survey/page.tsx
zz-hh-aa: Keen eye 👁️  nits all addressed / updated! Thanks Gabry :)

PR #508, opened 2025-06-03
feat: create new results page and reroute form on completion

zz-hh-aa:


app/results/page.tsx
DafyddLlyr, line 1: Our `/app` folder is quite messy with non-route folders, we should hoist these out back into `/src` to tidy this up - sorry for not spotting this sooner!

DafyddLlyr (changes requested):
Apologies it's taken me so long to get to this one!

PR #534, opened 2025-06-17
refactor: share custom components where possible

### LINKED ISSUE #526: Refactor: repetition in CustomLabelList props
Repetition emerged in [#510](https://github.com/theopensystemslab/fairhold-dashboard/pull/510#discussion_r2127215681) refactor

zz-hh-aa:
# What does this PR do?
- Renames `types.ts` to `shared.ts` (not strictly types anymore, nor utils)
- Creates a shared `CustomTick` component for coloured tenure labels and `BarLabelListTopLeft` for labels coloured by tenure that sit on the top left of bars
    - We do use custom LabelList components elsewhere, but their styling / formatting is different and it didn't make sense to combine

# Why? 
- Continuing the refactor started in #510 but I think my brain needed to do it incrementally! 

Closes #526

### DIFF
--- a/app/components/graphs/HowMuchFHCostBarChart.tsx
+++ b/app/components/graphs/HowMuchFHCostBarChart.tsx
@@ -9,7 +9,7 @@ import {
   StyledChartContainer,
 } from "../ui/StyledChartContainer";
 import { useState } from "react";
-import { BarLabelListTopLeft, CustomTick} from "./types"
+import { BarLabelListTopLeft, CustomTick, getLabel, getColor } from "./shared"
 
 const chartConfig = {
   freeholdLand: {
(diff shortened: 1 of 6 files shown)


app/components/graphs/HowMuchFHCostBarChart.tsx
```diff
@@ -136,23 +61,23 @@ const HowMuchFHCostBarChart: React.FC<StackedBarChartProps> = ({
 
   const chartData = [
     {
-      tenure: "freehold",
+      tenure: "Freehold",
```
zz-hh-aa, line 64: This is just for consistency across charts so we can share `getLabel` (though should have been consistent to begin with...)

app/components/graphs/shared.tsx
```diff
@@ -0,0 +1,113 @@
+import { formatValue } from "@/app/lib/format";
+
+export interface CustomLabelListContentProps {
```
DafyddLlyr, line 3: nit: Having both `?` and `| undefined` is redundant

app/components/graphs/shared.tsx
DafyddLlyr, line 1: As this file grows, we might find we actually do want a `types.ts`, `utils.ts` and `shared/labels/<COMPONENT_NAME.tsx`.

app/components/graphs/shared.tsx
DafyddLlyr: Ah sorry, I missed that previous comment. It might be the way tsconifg is set up in this project - thanks!

PR #543, opened 2025-06-24
style: restyle bar graphs

zz-hh-aa:
Updates styles for bar graphs that do _not_ need the data shape to change. 

Will do the more elaborate changes in further PRs (eg housing outcomes with ability to select tenure)
image

### DIFF
--- a/app/survey/components/SurveyGraphCard.tsx
+++ b/app/survey/components/SurveyGraphCard.tsx
@@ -8,7 +8,7 @@ type Props = React.PropsWithChildren<{
 const SurveyGraphCard: React.FC<Props> = ({ title, subtitle, children }) => {
   return (
     <div className="justify-center w-full bg-white m-4 p-4">
-        <h3 className={`text-xl md:text-lg sm:text-md font-bold text-black`}>{title}</h3>
+        <h3 className={`text-xl md:text-lg sm:text-md font-bold text-black`} dangerouslySetInnerHTML={{ __html: title }}/>
         {subtitle && <p className={`text-md md:text-lg text-gray-600 mt-2 font-normal`}>{subtitle}</p>}
       {children && <div className="mt-4 h-[calc(60vh-16rem)]">{children}</div>}
     </div>
--- a/app/survey/components/SurveyGraphCard.tsx
+++ b/app/survey/components/SurveyGraphCard.tsx
@@ -3,14 +3,20 @@ import React from "react";
 type Props = React.PropsWithChildren<{
   title: React.ReactNode;
   subtitle?: React.ReactNode;
+  action?: React.ReactNode;
 }>;
 
-const SurveyGraphCard: React.FC<Props> = ({ title, subtitle, children }) => {
+const SurveyGraphCard: React.FC<Props> = ({ title, subtitle, action, children }) => {
   return (
     <div className="justify-center w-full bg-white m-4 p-4">
         <h3 className="text-xl md:text-lg sm:text-md font-bold text-black">{title}</h3>
-        {subtitle && <p className={`text-md text-gray-600 mt-2 font-normal`}>{subtitle}</p>}
-      {children && <div className="mt-4 h-[calc(60vh-16rem)]">{children}</div>}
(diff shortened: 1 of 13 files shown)


app/survey/components/SurveyGraphCard.tsx
```diff
@@ -8,8 +8,8 @@ type Props = React.PropsWithChildren<{
 const SurveyGraphCard: React.FC<Props> = ({ title, subtitle, children }) => {
   return (
     <div className="justify-center w-full bg-white m-4 p-4">
-        <h3 className={`text-xl md:text-lg sm:text-md font-bold text-black`}>{title}</h3>
-        {subtitle && <p className={`text-md md:text-lg text-gray-600 mt-2 font-normal`}>{subtitle}</p>}
+        <h3 className={`text-xl md:text-lg sm:text-md font-bold text-black`} dangerouslySetInnerHTML={{ __html: title }}/>
```
DafyddLlyr, line 11: ⚠️ This has a scary name for a reason ([docs](https://react.dev/reference/react-dom/components/common#dangerously-setting-the-inner-html))

Why doesn't the original method work?

```ts
<h3 className={`text-xl md:text-lg sm:text-md font-bold text-black`}>{title}</h3>
```

app/survey/components/graphs/WhyFairhold.tsx
```diff
@@ -5,7 +5,12 @@ import { Bar, BarChart, XAxis, YAxis, ResponsiveContainer } from "recharts";
 import { useSurveyContext } from "../../context";
 
 export const WhyFairhold = () => {
-  const whyFairhold = useSurveyContext().barOrPie.whyFairhold;
+  let { whyFairhold } = useSurveyContext().barOrPie;
+  whyFairhold = whyFairhold.slice(0,5);
+  const maxValue = whyFairhold.length > 0 ? whyFairhold[0].value : 0;
+  const maxResponses = Math.ceil(maxValue / 10) * 10;
+  const xDomainMax = maxResponses < 10 ? 10 : maxResponses;
```
DafyddLlyr, line 12: This should be wrapped in a well-named function - without context it's hard to know what's going on here - it's also a repeated bit of logic.

Maybe something like this?

```ts

const getTopFive = (data: BarOrPieResult[]) => data.slice(0,5);

const calculateChartMaximum = (data: BarOrPieResult[]) => {
  const maxValue = data[0]?.value ?? 0;
  const maxResponses = Math.ceil(maxValue / 10) * 10;
  const chartMax = maxResponses < 10 ? 10 : maxResponses;
  return chartMax
}

```

app/survey/components/graphs/WhyNotFairhold.tsx
```diff
@@ -27,23 +31,34 @@ export const WhyNotFairhold = () => {
   }
   
   return (
-      <SurveyGraphCard title="Why wouldn't you choose Fairhold?">
-          <ResponsiveContainer>
+      <SurveyGraphCard 
+        title="Why <em>not</em> Fairhold?"
```
DafyddLlyr, line 35: Ah! This is why we need `dangerouslySetInnerHTML`.

I think the issue is just that we have `title` as a `string` type which means we get the literal `<em>` appearing in the title.

If we change the props of `SurveyGraphCard` to something like this...

```ts
type Props = PropsWithChildren<{
  title: ReactNode;
  subtitle?: ReactNode;
}>;
```

...and then just pass in something like this, I think we'll achieve what we're looking for safely.

```suggestion
        title={<>Why <em>not</em> Fairhold?</>}
```

This will work as `ReactNode` is a union which contains the `string` type, so `title` can be a chunk of JSX or just a plain string.

DafyddLlyr (changes requested):

app/survey/components/SurveyGraphCard.tsx
zz-hh-aa: Ah this was because the designs show an italicised header and this was my way of enabling rich text. 
image

Do you know of a safer way?

DafyddLlyr (approved):

PR #545, opened 2025-06-25
Oz/ranked tenure

zz-hh-aa:


app/survey/components/graphs/AnyMeansTenureChoice.tsx
```diff
@@ -1,20 +1,108 @@
 import React from "react"
 import SurveyGraphCard from "@/app/survey/components/SurveyGraphCard";
-import { CustomSankey } from "../CustomSankey";
-import { ResponsiveContainer } from "recharts";
+import { Bar, BarChart, XAxis, YAxis, ResponsiveContainer, Cell, LabelList } from "recharts";
 import { useSurveyContext } from "../../context";
+import { TENURE_COLORS } from "../../constants"
+import { CustomLabelListContentProps } from "@/app/components/graphs/shared";
```
zz-hh-aa, line 6: Dashboard graphs and survey results graphs are now using custom LabelLists. Should this be moved to another shared directory?

DafyddLlyr: Yep, I think only pages should be within `/app` and we should hoist these to `/components`

app/survey/components/graphs/AnyMeansTenureChoice.tsx
```diff
@@ -1,20 +1,108 @@
...
+                )}
+                  /> 
+              <Bar dataKey="value">
+                {anyMeansTenureChoice.map((entry, index) => {
+                    let answerStr = "";
+                    if (Array.isArray(entry.answer)) {
+                        answerStr = entry.answer[0] ?? "";
+                    } else if (typeof entry.answer === "string") {
+                        answerStr = entry.answer;
+                    }
```
DafyddLlyr, line 89: A few questions / thoughts on this - 
 - Is this actually an issue because we don't have strict / accurate enough types? Is this data genuinely a mixture of strings and arrays?
- We should probably handle this on the backend as part of the aggregation

app/survey/utils.tsx
```diff
@@ -78,25 +78,37 @@ const initializeSankeyResultsObject = (): SankeyResults => {
...
+            value.forEach((item, idx) => {
+                // Tidying the string
+                let shortAnswer = item.split("––")[0].trim();
+                shortAnswer = shortAnswer.replace("Freehold", "Market purchase")
+                // For now: 5 points for position 1, 4 for position 2, etc. 
+                const points = Math.max(5 - idx, 1);
+                addResultItem(results.anyMeansTenureChoice, shortAnswer, points);
+            });
+            return;
+        }
```
DafyddLlyr, line 109: nit: There's enough logic and complexity here for this to be a helper function.

app/survey/utils.tsx
```diff
@@ -78,25 +78,37 @@ const initializeSankeyResultsObject = (): SankeyResults => {
...
+
+        // We also handle housingOutcomes separately since we are separating output graphs by currentTenure
         if (key === "housingOutcomes") {
```
DafyddLlyr, line 112: Do we need to also exclude `"anyMeansTenureChoice"` here?

DafyddLlyr (changes requested):

app/survey/components/graphs/AnyMeansTenureChoice.tsx
zz-hh-aa: Good catch, yeah types could have been tighter! They now are in [56800ca](https://github.com/theopensystemslab/fairhold-dashboard/pull/545/commits/56800cadb02783f220aa5a41202e1304eef0c5bb)

DafyddLlyr: Ah a simpler solution that I thought - nice!

DafyddLlyr (approved):
Approved following dev call ✅ 

Plan of action - 
- Merge this PR and associated (branched) ones
- Revisit type question and tidy this up in a follow up

PR #546, opened 2025-06-25
fix: use client side rendering

zz-hh-aa:
I noticed this hydration error for the first time this afternoon (on `main`--not just feature branches, only on Chrome). 
Screenshot 2025-06-25 161423

Is it a bad idea to render client-side instead? This fix addresses the issue by showing load state until mounting is finished in client (and so gets rid of the hydration mismatch), if I understand correctly.

### DIFF
--- a/app/components/ui/CalculatorInput.tsx
+++ b/app/components/ui/CalculatorInput.tsx
@@ -1,5 +1,5 @@
 "use client";
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import { useForm } from "react-hook-form";
 import { zodResolver } from "@hookform/resolvers/zod";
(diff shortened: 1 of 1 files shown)


DafyddLlyr (commented):
I actually can't recreate this hydration error locally (and I'm not sure what would be causing it now).

Are you still getting this problem on `main`?

zz-hh-aa:
> I actually can't recreate this hydration error locally (and I'm not sure what would be causing it now).
> 
> Are you still getting this problem on `main`?

Yes I'm still getting it ! Only on Chrome--138.0.7204.97 on Windows 11. I've also tested on Microsoft Edge, Firefox, Arc, Brave and only Chrome shows the error 🤔

PR #563, opened 2025-07-01
feat: add footer to survey results

zz-hh-aa:
Moved `Header` into `components/` to share with survey results.


app/survey/page.tsx
DafyddLlyr, line 1: a11y: HTML `header` and `footer` should be outside the `main` component, not nested within it.

PR #593, opened 2025-07-18
feat: new route for calculator results v3

### LINKED ISSUE #502: Create new route to enable saving form fields
Form at `index`, we have a conditional query. On submit, redirect to `results`. 

1. Can I use Next to create a random page eg "Hello world"
2. Link component / function, send to `/results`, on submit, save data to state
3. Take form data and pass as prop to `/results` or maybe another internal state
4. If no data, redirects to home

zz-hh-aa:
This is a new version of #508 (closed because of messy rebase and to improve separation of concerns) _and_ #580 (because rebasing got too messy). 

# What does this PR do?
Creates a new route at /results for survey dashboard
Moves conditional rendering to /results/page.tsx (it probably shouldn't have been in CalculatorInput.tsx anyway!)

# Why?
We decided to create a new route for results so that if people hit 'back' to change any of their form inputs, they wouldn't get redirected to their previous non-Fairhold page / referrer.

# Questions
I was thinking about [Daf's comment here](https://github.com/theopensystemslab/fairhold-dashboard/pull/508#discussion_r2177813846) and wondered if maybe we would want to route people to `/results?housePostcode=abc123`--eg I want to share with my flatmates / partner what our home would cost if it was Fairhold, or similar. Thoughts? 

Closes #502

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -1,16 +1,53 @@
+"use client";
+import React, { useState } from "react";
 import CalculatorInput from "../components/custom/ui/CalculatorInput";
 import { ClipLoader } from "react-spinners";
 import { Suspense } from "react";
 import { Footer } from "../components/custom/ui/Footer";
 import { Header } from "../components/custom/ui/Header";
+import { FormFrontend } from "@/schemas/formSchema";
+import { useRouter } from "next/navigation";
 
 export default function Home() {
+  const router = useRouter();
+  const [error, setError] = useState<string | null>(null);
+  const [isSubmitting, setIsSubmitting] = useState(false);
+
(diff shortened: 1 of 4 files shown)


zz-hh-aa:
> I am not sure if I get right your question with respect to Daf's comment, maybe we can chat at the first chance?

Thanks Gabry! I think Daf was saying that we most likely wouldn't want to send users directly to a results page (eg by giving them a URL with already input params `/results?housePostcode=abc123`), but I was thinking this might be useful (eg if I want to show a partner, flatmate etc. how much our home would cost under Fairhold). So I've left the functionality in for now, can def discuss again!

PR #607, opened 2025-07-22
style: update page layouts to match webflow

zz-hh-aa:
Update page width to match webflow. First we tried max container with 1280 with left 1/4 column empty, which looked strange. Now all content is in a max 960px content block, centred. 

Webflow example:
Screenshot 2025-07-22 161241

1280 width layout with 1/4 and 3/4 cols:
Screenshot 2025-07-24 093011

Now:
Screenshot 2025-08-05 134446

### DIFF
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -40,6 +40,7 @@ export default function Home() {
   return (
   <div className="min-h-screen w-full">
     <Header />
+    <div className="hidden md:block top-spacer"/>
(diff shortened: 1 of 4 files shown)


zz-hh-aa:
Ah great catch, thanks! I've added an `m-5` for any screen under `lg`. 

Screenshot 2025-08-08 093209

Screenshot 2025-08-08 093155

DafyddLlyr (approved):

PR #620, opened 2025-08-06
style: survey results font fixes

zz-hh-aa:
This PR makes various small fixes to bring text closer to designs. Updates some heading hierarchy, sizes and colours; and moves / shares the `Highlight` component for reuse.

### DIFF
--- a/components/custom/dashboard/cards/WhatDifference.tsx
+++ b/components/custom/dashboard/cards/WhatDifference.tsx
@@ -1,11 +1,11 @@
 import { Household } from "@/models/Household";
-import { Drawer } from "../../ui/Drawer";
-import GraphCard from "../../ui/GraphCard"
+import { Drawer } from "@components/custom/ui/Drawer";
+import GraphCard from "@components/custom/ui/GraphCard"
 import { DEFAULT_INTEREST_RATE, DEFAULT_MORTGAGE_TERM, DEFAULT_INITIAL_DEPOSIT } from "@/models/constants";
 import ReactMarkdown from 'react-markdown';
 import explanationContent from '../help/WhatDifference.md';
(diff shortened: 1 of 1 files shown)


components/custom/survey/graphs/Age.tsx
```diff
@@ -27,7 +27,11 @@ export const Age = () => {
                             <Cell key={`cell-${index}`} fill={COLORS[index % COLORS.length]} />
                         ))}
                     </Pie>
-                    <Legend align="center" verticalAlign="bottom" />
+                    <Legend 
+                        align="center" 
+                        verticalAlign="bottom" 
+                        wrapperStyle={{ fontSize: 18 }}
+                    />
```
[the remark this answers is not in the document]
DafyddLlyr: I think this is fine! If we find we're doing it again I'd look at theming these or making variants.

components/custom/dashboard/cards/WhatDifference.tsx
```diff
@@ -5,6 +5,7 @@ import { DEFAULT_INTEREST_RATE, DEFAULT_MORTGAGE_TERM, DEFAULT_INITIAL_DEPOSIT }
 import ReactMarkdown from 'react-markdown';
 import explanationContent from '../help/WhatDifference.md';
 import { DEFAULT_FORECAST_PARAMETERS } from "@/models/ForecastParameters";
+import Highlight from '../../ui/Highlight';
```
DafyddLlyr, line 8: Could we also use the alias `@compoents` here? Maybe we should find an eslint rule to enforce the use of these.

DafyddLlyr (approved):

components/custom/dashboard/cards/WhatDifference.tsx
zz-hh-aa: Good shout!

PR #649, opened 2025-08-26
style: mobile sankeys

zz-hh-aa:
Uses `useState` to set different margins, padding and font size for sankeys on mobile. 

Before:
Screenshot 2025-08-26 132020

After:
Screenshot 2025-08-26 132012

Before:
Screenshot 2025-08-26 131949

After:
Screenshot 2025-08-26 132002


gabrielegranello (approved):
At the beginning I was confused why you were using useEffect only to set the margins only once (when the app is mounted) instead of simply coding it. However I can see that `window` is not available if the page is rendered on the server and that you need useEffect to actually call `window`. Clever!

PR #650, opened 2025-08-26
chore(wip): rewrite styles using tailwind only

zz-hh-aa:


app/globals.css
```diff
@@ -193,4 +194,31 @@ main {
 .scrollbar-hide {
   -ms-overflow-style: none;  /* IE and Edge */
   scrollbar-width: none;     /* Firefox */
-}
\ No newline at end of file
+}
+
```
zz-hh-aa, line 198: I copied these over from webflow since they had a few more properties, but is this kind of defeating the purpose of replacing webflow styles? Should they just be written in-line with Tailwind?

components/custom/ui/Footer.tsx
```diff
@@ -1,80 +1,81 @@
...
-          </a><a href="#" className="footer-link-block w-inline-block">
-            <div className="body-s">Accessibility</div>
-          </a><a href="#" className="footer-link-block w-inline-block">
-            <div className="body-s">Contact
-              us</div>
-          </a>
-          <div className="spacer"></div>
+  <footer className="bg-[rgb(var(--footer-background-rgb))] text-white py-8 ">
+    <div className="max-w-[1280px] mx-auto px-4">
+      {/* ROW 1 */}
```
zz-hh-aa, line 6: Is it okay to leave these in? Found they really helped for readability, kept getting lost in the divs...

PR #666, opened 2025-09-18
feat: ankle and footer

zz-hh-aa:
Rebuilds the footer and creates an 'ankle' for the survey. 

- Alastair decided [we shouldn't include socials in footer](https://opensystemslab.slack.com/archives/C01RF806126/p1758116795508909?thread_ts=1752674008.051099&cid=C01RF806126) since we don't really use them
- ^ and also to use a 1px container-width divider between ankle and footer
- We also [decided to remove webflow CSS](https://github.com/orgs/theopensystemslab/projects/5?pane=issue&itemId=93532270&issue=theopensystemslab%7Cfairhold-dashboard%7C282) and just use tailwind, so the footer is rebuilt

Looks like this:
image


components/custom/ui/Footer.tsx
DafyddLlyr, line 1: A good alternative to the `{/* ROW 1 */}` comments here would be to wrap them up as components - 

```ts
<RowOne>
  <Col1/>
  <Col2/>
  <Col3/>
</RowOne>
<RowTwo>
  <Col4/>
  <Col5/>
  <Col6/>
</RowTwo>
```

This still isn't _that_ meaningful, but it's a little easier on they eyes maybe?

I also really like use CSS grid template areas which are a nice way of imparting semantic meaning. IDs can also be a good solution here.

DafyddLlyr (approved):
Looking great!

components/custom/ui/Footer.tsx
zz-hh-aa: `grid-template-areas` is so cool! Hadn't encountered before :) 

I had a play around with a couple options here--seemed redundant to have multiple numbered empty block columns (eg for `<div className="hidden md:w-1/4 md:block" />`) but also odd to have inconsistent block numbers (eg `Col1`, `div`, `div`, `Col2`). Going to keep as is for now!

PR #731, opened 2026-03-06
feat: monthly rent weighted by bedrooms and dwelling type

zz-hh-aa:
We are now using a different rental dataset and different method of calculating monthly rent, which takes bedrooms and house type into account. This PR updates the schema, data fetching and calculations. 

~~We won't need to truncate the rent table on production since the migration `DROP`s it, we will need to manually insert the data though (which we'll be doing as part of the data update anyway). The order of operations here will be to merge this, then run all the SQL changes on prod directly (trialled in my local db first)~~

I think the sequencing here should actually be:
1. Drop table by running SQL directly on prod
2. Merge this PR to change Prisma schema
3. INSERT directly on prod


DafyddLlyr (approved):
Apologies for the delay here...!

If we merge this, the migration will empty and restructure the `rent` table (`DROP` + `CREATE`). The Prisma schema is updated to show the new structure. I think all of this looks ok and should work as expected, we just need to then handle `INSERT` manually as you say.

I guess this works locally without issue?

We could be a bit more robust here and rename `rent` to `rent_old`, then just insert a `rent` table, and then delete `rent_old`, but I think that just gets us to the same exact place with more steps/work?

prisma/migrations/20260306171718_update_rent_to_new_shape/migration.sql
```diff
@@ -1,24 +1,3 @@
-DROP TABLE IF EXISTS "rent";
```
zz-hh-aa, line 1: Dropping manually, so removed from here

zz-hh-aa:
Ultimately going with this sequencing!

1. Manually rename `alter table rent rename to rent_old`
2. Merging this and running migration, thus creating new `rent` table
3. Manually insert to populate (along with running the rest of the sql on prod to finish data update)
4. Test!
5. Drop `rent_old`

